Issue #8, opened 2025-06-30
Small issues


pyrareae:
I have been testing this a bit, amazing work! I just wanted to write down some observations on some rough edges from a sort of QA PoV in case that is helpful. I might try to take care of some of these myself if contributions are welcome! These could probably be listed as separate issues.

I am running this in Arch Linux (well, Arch in a distrobox on Aurora, to be precise) on an Intel based laptop, so of course more things are going to break more here lol.

### Grain glitch

When I activate the film grain effect I this this off diagonal artifact. This might be a driver e on my system.

![Image](https://github.com/user-attachments/assets/9c939d06-7167-4d9e-a31c-0d296e00e4ff)

### Background transparency

I'm not sure if blur is activated on osx and windows, but on linux the app is translucent but has no background blur. Other apps like the terminal are translucent and blurry. I believe there is a flag you can pass to the window manager to activate this.

![Image](https://github.com/user-attachments/assets/c701d6b9-145c-4007-8833-3d9768632c47)

### Debouncing

When I move any of the controls it feels like it's being passed through a denouncing function. and I have to wait about .5s to see the change happen. This feels a little "laggy" even for adjustments that are usually pretty fast.

### Zooming

Zooming origin is on the corner and not the center of the image, this is a bit awkward. Also 100% seems to be meaning "fit" rather than 1:1 pixel ratio as with most other software.

CyberTimon:
Hello @pyrareae,

Thanks a lot for your detailed feedback and report!

First off - yes, contributions are absolutely welcome. I’m currently juggling this with my apprenticeship, so any help is super appreciated.

Let’s go through your points one by one:

### 1. Grain glitch
This issue is likely GPU-related - I haven’t been able to reproduce it on my end. I’ve tested on an RTX 4070 and a MacBook Pro, and everything looked fine.
Fixing the grain isn’t super high on my list right now, but it also shouldn’t be too difficult since it’s just a simple part of the shader code. Maybe AI can help you tweak it for your Intel-based setup.

### 2. Background transparency
Yep, I’m aware of this one. On macOS and Windows, the window blur gives a nice polished look. But on Linux, the windows are fully transparent, which can be kind of distracting.
I’ve been using [windows-vibrancy](https://github.com/tauri-apps/window-vibrancy), which sadly doesn’t support Linux. I haven’t found a good workaround yet - mainly because I haven’t had time to dig deeper.
Might be a pretty easy fix with some tweaks in the main.rs file, though.

[...]

PR #14, opened 2025-07-02
fix(linux): set window transparency false

andrewazores:
See #8

The point of this is just to disable window transparency on Linux, where the `window-vibrancy` crate doesn't support actually applying any such window decorations like blur effects and just ends up with translucent, unblurred, hard-to-use windows. It would be better to just render opaque windows.

Initially I simply applied a patching step in the CI action to set the JSON property to `false` for Linux, but then I figured it's likely that in the future there may be more platform-specific patches that come up and which cannot use Rust macros like `#[cfg(target_os = "windows")]` (because they are not in Rust source files). I used the common Linux `find` utility to put something together which allows applying `.patch` files using the `patch` command, or else dynamically applying patches by executing executable files like the one here which runs `yq`. For MacOS there should be an equivalent to `find` already available, but I don't know what would work for Windows.


andrewazores:
@pyrareae would you be able to test this change and see if it helps on your system? For me, with the window manager/compositor I use I already don't have window transparency support, so I don't see any change.

pyrareae:
I'll take some time to test this soon!

andrewazores:
If you have `yq` on your system you can just run `sh patches/linux/0001-window-transparency.sh` from the project root directory. If not, then just manually edit the `src-tauri/tauri.conf.json` and set the window transparency property to `false` manually. Then `npx tauri dev --release` or `NO_STRIP=true npx tauri build` and run the application.

andrewazores:
FWIW I just manually tested it on another system running Gnome Shell.

Before patch:
Screenshot From 2025-07-05 12-28-26

After:
Screenshot From 2025-07-05 12-30-15

It's definitely not as pretty as the Mac/Windows screenshots, but I think it ends up more usable this way.

andrewazores:
@pyrareae @CyberTimon ready for your review now :-)

andrewazores:
For future improvement, there is also a way to set window transparency via Tauri API:

https://github.com/tauri-apps/tauri/issues/7433

So, it would be nice to have something like a menu bar, with a menu like `Window > Preferences`, and a toggle for the window transparency effect, and blur effect. Maybe the default values for this can be based on the platform, but the user can still have their choice.

CyberTimon:
Thank you @andrewazores for creating this PR and thank you @pyrareae for testing it. 

The PR looks fine and does exactly what we want to achieve (until [windows-vibrancy](https://github.com/tauri-apps/window-vibrancy/tree/dev) provides a better alternative).

I will merge this now :)

PR #16, opened 2025-07-02
ci(upload): upload built unreleased artifacts

andrewazores:
When the non-release CI actions complete (on PR, or on merge to `main`), this uploads the built artifacts so that contributors/testers/over-eager-users can directly download and install the binaries corresponding to that changeset - something like nightly builds. This will make it quicker and easier for the community to help test development builds without needing to know how to actually build the project from source themselves (or be on a supported Linux distribution where that's even feasible).

See https://github.com/andrewazores/RapidRAW/actions/runs/16032986502 for an example of how the builds appear.


CyberTimon:
Thanks for this. I will check it out soon. Quick question: is it possible to also upload the macos / windows binaries seperate, like tauri action does already?

andrewazores:
You mean so that the dmg and exe are "bare" files like the rpm/deb/AppImage, instead of zipped? I can do that.

CyberTimon:
Yes exactly. On the current build workflow, which runs on release and works with tauri actions, it already does this. Would be cool to keep similar.

andrewazores:
New run:

https://github.com/andrewazores/RapidRAW/actions/runs/16034708187

image

Annoyingly, despite the filenames being ex. `RapidRAW-sha.x86_64.rpm`, when you download the file through the GitHub Actions UI you will get `RapidRAW-sha.x86_64.rpm.zip`, as documented by the action/helper script I used: https://github.com/lhotari/gh-actions-artifact-client?tab=readme-ov-file#usage-tips

But the file is not actually zipped. Renaming the file to simply remove the `.zip` allows me to directly install the package file.

This is related to https://github.com/actions/upload-artifact/issues/426 , I think.

Another possible solution could be to run the `release` Action on a cron job, ex. daily (to produce actual nightly builds), and have those consistently upload to a "nightly draft release". This doesn't allow for direct testing of individual PRs or commits, but it might be another useful avenue for testers.

Anyway @CyberTimon I think I've done as much as I can for this PR for now.

CyberTimon:
Thank you a lot for taking your time contributing to this. 

I will definitely take a look at this tomorrow or on friday. I was implmeneting rawler the whole evening so I couldn't look at this in detail yet. 
Rawler actually works, I just have some bugs with the color profile and white balance.

andrewazores:
At one point while working on this I had it so that the uploaded file *was* actually a zip archive, so at least when the UI downloads the file and names it that way, it's correct. Users could then actually unzip the archive and get the (one single binary) file they wanted. I removed that because it worked on MacOS and Linux, but not on Windows, where the `zip` command isn't available so the CI job just failed.

Maybe I should bring that back, and find a way to make it work on Windows as well. This should be less confusing than how it is now, with the `.zip` suffix being a lie.

andrewazores:
Note for myself to follow up on that:

https://github.com/marketplace/actions/zip-release

https://mysticmind.dev/how-to-create-a-cross-platform-zip-archive-using-github-actions

andrewazores:
Currently the MacOS and Linux downloads are actually zipped and get downloaded with the `.zip` suffix, so that makes sense. Windows downloads will see `.exe.zip`, but it isn't actually a zip archive, so users would need to rename it to remove the `.zip` before being able to run it - or else just wait for an actual release, where the artifact will be handled as usual via the `tauri-action`.

Fixing the CI artifact upload specifically for Windows is getting to be annoying, and I'm tempted to say just leave it as-is with a documentation note until enough people complain...

andrewazores:
Agh, hang on. At some point here I've now made it so that when you download the zipped files, you don't just get the file in a zip archive, you get a directory chain:

```
unzip RapidRAW-c5b736312b1254edb10b5bb0ca2539135ad0bf6c.x86_64.rpm.zip 
Archive:  RapidRAW-c5b736312b1254edb10b5bb0ca2539135ad0bf6c.x86_64.rpm.zip
  inflating: src-tauri/target/release/bundle/rpm/RapidRAW-1.1.1-1.x86_64.rpm
```

andrewazores:
New run, should be fixed. Windows "false .zip" issue is still there.

https://github.com/andrewazores/RapidRAW/actions/runs/16042069014

CyberTimon:
Hey @andrewazores 

I just finished working on the rawler implementation (switch rawloader -> rawler). This should add support for your camera, so you might finally be able to actually use the programm with your camera. Can you please give me a short feedback if this works now?

Regarding this PR (and the PR based on this, intel mac's), are you still trying to fix the windows false .zip issue or is this ready to merge? I'm fine if the binaries in the commit artefacts are a bit broken, but on release they should be a normal installer, like before. 

Thank you!

andrewazores:
I can file an issue to track the Windows false zips, and maybe come back to it later on at some point. For now I'm done with it. The full release should still work as it did before.

andrewazores:
@CyberTimon rawler is working great now, I can see my NEF files perfectly :tada:

CyberTimon:
Thank you. With the issue it's all good for me. I will merge this - we can work on this in the future, after all the core priorities are done.

CyberTimon:
> @CyberTimon rawler is working great now, I can see my NEF files perfectly 🎉

Amazing - thanks for testing! 💯 
I actually wasted so much time trying to get the correct color profiles etc for rawler while the library already does this all by itself hahah. It sadly doesn't have a good documentation..

Issue #17, opened 2025-07-02
Set up CI caching


andrewazores:
Currently, every CI run starts from a completely clean slate, and all dependencies (Rust crates) need to be downloaded and compiled. Since these versions are pinned, the output from a download and compile cycle should be reusable for the next time a build is run, saving time and compute.

https://github.com/Swatinem/rust-cache looks like it should help.

PR #18, opened 2025-07-02
ci(cache): add rust-cache

andrewazores:
Fixes https://github.com/CyberTimon/RapidRAW/issues/17

See https://github.com/andrewazores/RapidRAW/actions/runs/16034816728 for a CI run before the cache was enabled. This takes ~13 minutes to complete currently - each OS is done in parallel, so the slowest (Windows) defines the overall time.

See https://github.com/andrewazores/RapidRAW/actions/runs/16035873648/job/45247392134 for a CI run with cache primed. The build time is reduced to 7 minutes. This one also includes #16 which is still a WIP and failed for Windows in this case, but for reasons unrelated to the cache.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,6 +33,8 @@ jobs:
 
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: src-tauri
 
       - name: Set up Node.js
         uses: actions/setup-node@v4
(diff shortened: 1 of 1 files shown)


andrewazores:
@CyberTimon this one is also ready

andrewazores:
Check failure just be unrelated. The workflows run on `pull_request_target` and use the definition from the current `main` branch, *not* the PR contents. So the workflow update here did not actually affect the run. I don't have permissions to rerun the action without just pushing a new commit.

CyberTimon:
Thanks a lot for this PR. Very cool improvement and I see nothing against merging this.

PR #23, opened 2025-07-06
feat(window): use tauri-controls for native-looking window controls

andrewazores:
image

Adds a dependency on https://github.com/agmmnn/tauri-controls to provide native-looking window controls.

The GNOME controls for Linux look a lot like Windows controls to my eye.

I think it would be nice to have some application settings for the user to choose if they want transparency effects, blur effects, and native-style window controls. I will continue experimenting with this PR to add these options to the existing settings menu.


andrewazores:
There is currently an issue preventing automatic platform detection from working:

```
Unhandled Promise Rejection: Command os_type not found
```

(this appears in the webview dev console)

This comes from the new `window-controls` dependency, but I'm not really sure what's going wrong there. I understand that there is an IPC message passing system between the React frontend and the Rust backend, and that this `os_type` command should be one of the IPC commands available, but I don't know why it isn't found.

Manual selection/override of the controls works just fine however. The default setting is to retain current app/RapidRAW-drawn controls. If platform detection fails it seems to fall back to Windows-style.

Screenshot_2025-07-07_16-56-27
Screenshot_2025-07-07_16-56-11
Screenshot_2025-07-07_16-55-50
Screenshot_2025-07-07_16-55-32

The `Native` setting using the tauri-controls `WindowTitlebar` component, which is supposed to handle not just rendering the controls in a way that visually matches the platform, but which should also handle laying out the titlebar according to the platform - ie MacOS controls on the left, else on the right. This also does not seem to be working as expected, at least not on my Linux system using i3/Sway or Gnome Shell, so if it also doesn't work as intended on Windows and MacOS then the last commit in this PR should be reverted.

andrewazores:
@CyberTimon could you test this on Windows/Mac and let me know if I should revert that last commit? And if you have any idea about the `os_type` command?

CyberTimon:
I haven't had time to test this yet but a quick search led me to this: https://github.com/agmmnn/tauri-controls/issues/37

Will test this in the evening and report my findings back to you - thanks!

andrewazores:
Yea, it seems like `tauri-controls` was updated to some point in the tauri v2 beta or RC phase, but things in tauri changed and tauri-controls stopped getting updates, so it's a bit broken. Luckily it seems that it's only the automatic platform detection that is broken and manually setting it works fine.

andrewazores:
Screenshot_2025-07-08_20-33-25

Ready for review now @CyberTimon . I dropped the native detection bits, so now the user must manually select the style they want. MacOS mode puts the window controls on the left side, otherwise they stay on the right side and just the appearance of the controls changes.

CyberTimon:
Thanks for the PR @andrewazores. I will look into this soon :)

CyberTimon:
This has some merge conflicts now. Are you ready with this or do you directly want to integrate / try to implement native controls by default? I will wait a bit until I merge this, as I want to focus on other priorities (such as faster thumbnail generation etc)

andrewazores:
Screenshot from 2025-07-10 09-21-48

Linux GNOME with transparency disabled and native window controls enabled.

@CyberTimon I haven't figured out how to remove the top grey bar when in native controls mode, the one that normally contains the app-rendered title text and controls. I guess it's just some css styling that needs to be tweaked. I can revisit this later on, or else if you know how to remove it when in native controls mode then please go ahead and update this PR (or commit a separate patch after).

src-tauri/src/main.rs
```diff
@@ -145,6 +146,7 @@ impl Default for AppSettings {
             editor_preview_resolution: Some(1920),
             sort_criteria: None,
             theme: Some("dark".to_string()),
+            controls: Some("app".to_string()),
```
andrewazores, line 149: Oh, also, I left the RapidRAW custom styled controls as the default setting here.

Just change this to `"native"` instead of `"app"` if you think it should default to the native controls. Personally, I think the custom styled ones look pretty nice anyway, but I think there should be the option. In my case, with my usual window manager, my application windows all have very thin titlebars with only the title text, and there are no control buttons at all, so native mode fits in better. But I know that setups like mine are by far a tiny minority.

CyberTimon:
Hey

I havent really looked into this yet, so my question might be already implemented.:
Would it be possible to make the app's default controls (how they are right now) on WIndows and do the custom (os specific) controls on macOS / Linux? 

I plan to review it this evening. 

Thanks

andrewazores:
You mean to make those the defaults? As it is now, it defaults to doing the same thing you originally implemented ("app" controls) on all platforms, but the user can pick from any of the other options that I added.

My suggestion if changing that scheme would be to continue using the "app" controls on Windows and use the new native controls functionality on MacOS and Linux. The "app" controls are Windows-esque anyway so it fits in well there and it's relatively common for Windows applications to do this kind of styling. On MacOS and Linux it's very rare for applications to have this kind of custom styling, so just directly using native window decorations/controls probably makes the most sense.

But then in that case, the grey bar should be removed from native mode before making that the default anywhere.

CyberTimon:
Yes, I agree with you. For native controls, I believe the grey bar is caused by padding in both the MainLibrary and the main app. If you’re able to implement the desired behavior - where the control remains selectable but is automatically the “correct” one - I can definitely remove the grey bars. Shouldn't be too hard to fix :)

andrewazores:
Last couple of commits should do the job I think. Probably not the cleanest way. This also sets the behaviour back to defaulting to transparency disabled on Linux.

andrewazores:
I guess some time later on when I have more time to look at #38 and the runtime OS version detection, I can also come back to this and refactor the OS type detection using that dependency to clean up the logic I just added.

CyberTimon:
@andrewazores  Did you also run into this issue? https://github.com/agmmnn/tauri-controls/issues/28

andrewazores:
I did when I followed tauri-controls README. But some digging through reveals that tauri-controls actually doesn't depend upon tauri-plugin-window anymore. Just ignore that instruction and remove the dependency.

CyberTimon:
Yes - I'm testing some things. Will report my findings back to you soon.

CyberTimon:
Okay so thanks for all the help and digging into tauri-controls! 

I ended up finding a much simpler way to handle this directly in the frontend, which I've just pushed. It uses the Tauri OS plugin for detection, so we can avoid the broken dependency and the need for a user setting. I don't really like that we make it a user setting.

Could you give it a try on Linux? It currently falls back to Windows-style controls, so it would be great if you could add a Linux-specific section to the component.

Commit which includes this: https://github.com/CyberTimon/RapidRAW/commit/b72d75c6218129290373f75faeb37e595207e55b

Still thanks for your PR, even though I don't merge it. I will soon add all the @src pointers to the main branch, they are good 👍

andrewazores:
Your commit still ended up using `tauri-controls`, same as mine...

I would still suggest considering implementing the work I did here to use the native `.decorations()` function on the backend. Doing this purely on the frontend using React components does not actually allow for native adaptability of the UI - Linux does not have a unified UI theme like MacOS and Windows do, so it's impractical (even impossible really with all the custom theming possibilities) to use pure frontend styling like you have done and make sure it always matches the user's environment.

If you want, I can rework this to make it not a user setting and just do it automatically. I really, really think that using the `.decorations()` method I implemented, even though it's more complex, is the only feasible way to make this feel like a cohesive UI experience for Linux users.

CyberTimon:
@andrewazores 

> Your commit still ended up using tauri-controls, same as mine...

Yes - sorry for the confusion. With the "simpler solution" I meant to not use the .decorations() method. I now start to realize why it could be better, especially in the linux area. I wasn't aware of all the different linux styling. 
If we are going to implement .decorators(), we probably have to remove all the padding in the App.jsx, which handles the MainLibrary and main Editor (app) top border.

Thank you a lot - I just saw that you created a PR, which I will take a look at soon. I need to study for a bit but then I will come back to this :)

andrewazores:
> Yes - sorry for the confusion. With the "simpler solution" I meant to not use the .decorations() method. I now start to realize why it could be better, especially in the linux area. I wasn't aware of all the different linux styling.

No problem, but next time please check with me (or any contributor) and ask questions about why changes were done particular ways before rewriting something else. As you see, I had my reasons :-)

CyberTimon:
Thank you. Yes, I'll remember that in the future.

PR #29, opened 2025-07-08
build(rpm): add Fedora and OpenSUSE .rpm dependencies as 'recommends' (weak/optional dependencies)

### LINKED ISSUE #27: Linux packages should pull in required dependencies, ex. webkitgtk
> Installed from the rpm and had to install 'libwebkit2gtk-4_1-0'. It works. I'll try it, even if I already noticed that RAF files are not rendered correctly.     
>  

 _Originally posted by @fabiosirna in [#20](https://github.com/CyberTimon/RapidRAW/issues/20#issuecomment-3032455918)_

https://v2.tauri.app/distribute/rpm/#setting-the-conflict-provides-depends-files-obsoletes-desktoptemplate-and-epoch

@CyberTimon I'll look into this ASAP.

andrewazores:
Fixes #27

Rather than setting up two separate builds for Fedora-family and OpenSUSE-family distributions, just to specify the slightly different package names for `webkit2gtk`, this simply uses RPM's "recommends" facility. This specifies "weak" or "optional" dependencies, which will be installed by default *if available*. If not available, or if installing the dependency would cause some other error (ex. package version conflict), then the dependency is simply skipped (at least by `dnf` in the Red Hat distro family, but I believe OpenSUSE's `zypper` behaves the same). This way, a single `.rpm` build should be able to work for both distribution families, and pull in the relevant webkit2gtk package as needed.


andrewazores:
@fabiosirna could you check that the RPM generated by CI here pulls in the correct webkit2gtk package on your Tumbleweed system, and runs as expected?

https://github.com/CyberTimon/RapidRAW/actions/runs/16156700774?pr=29

CyberTimon:
Thanks for this PR. Looks fine - I'll merge this.

Issue #13, opened 2025-07-02
When moving the windowed editor around it moves incredbly slowly


Gormadt:
When attempting to move the editor around it moves along my window incredibly slowly not in sync at all with the mouse.

https://github.com/user-attachments/assets/4cdde748-57ab-4b60-ab10-a1587e651ed8

EDIT: I should mention that my current system contains a 7900XTX, 128GB of Ram, and an AMD Ryzen 7 5800X CPU. The system is also running Windows 10.

CyberTimon:
Hello @Gormadt 

Thanks for the report! 
This is a known Windows 10 bug and this also happens on applications. 

References:
https://github.com/Seo-Rii/electron-acrylic-window/issues/6
https://github.com/tauri-apps/window-vibrancy/issues/47
https://www.google.com/search?q=windows+acrylic+lag

It looks like it might be related to the mouse pooling rate. I might just provide an option in the near future to disable the blur (in the RapidRAW settings). 

Kind regards, 
Timon

Gormadt:
Thank you for the quick feedback, I'd love a setting to be able to disable it in the future.
That way it leaves the option for people to still use it and also the option for people who don't want to use.

CyberTimon:
Should be solved. You can now disable window transparency as https://github.com/CyberTimon/RapidRAW/pull/36 got merged. Feel free to reopen if you still encounter these bugs, even when using the newest release (which should get released this evening).

PR #36, opened 2025-07-09
feat(transparency): window transparency can be toggled in settings menu

andrewazores:
Related to #13
Related to #8
See also #14
Includes #44 , since a very similar change is also required to get this to work

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,22 +1,22 @@
 {
-  "name": "lightroom-clone",
-  "version": "0.1.0",
+  "name": "rapidraw",
+  "version": "1.2.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
-      "name": "lightroom-clone",
-      "version": "0.1.0",
+      "name": "rapidraw",
+      "version": "1.2.1",
       "dependencies": {
         "@tauri-apps/api": "^2.5.0",
         "@tauri-apps/plugin-dialog": "^2.2.2",
         "@tauri-apps/plugin-opener": "^2",
+        "@tauri-apps/plugin-process": "^2.3.0",
         "framer-motion": "^12.18.1",
         "konva": "^9.3.20",
         "lodash.debounce": "^4.0.8",
         "lucide-react": "^0.515.0",
         "react": "^18.3.1",
-        "react-beautiful-dnd": "^13.1.1",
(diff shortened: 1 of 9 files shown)


andrewazores:
@CyberTimon I tested this on Linux (Gnome) and it's working as intended, but please double check on Windows and MacOS since I've modified how the application window gets created. The [Tauri WindowBuilder ](https://docs.rs/tauri/latest/tauri/window/struct.WindowBuilder.html#known-issues) has a warning about window creation on Windows (ha), so I've followed the suggested workaround, but haven't verified it.

https://github.com/user-attachments/assets/8c007824-eee5-4763-88dc-c5a4d5faf6d3

The window configuration transparency setting (`false` on Linux, `true` for Windows/MacOS) should be picked up as the initial default value if the application settings do not have a value for it. The user can then override this preference and it will be persisted and used for each subsequent launch.

andrewazores:
Also, based on this work and #23, I think I can pretty easily build on top to add a proper "native controls" setting like what I tried to do with `tauri-controls` in #23. Once both of these are merged I will try that next.

CyberTimon:
On Windows it always crashes when changing the transparency:
```
[0710/123330.481:ERROR:ui\gfx\win\window_impl.cc:124] Failed to unregister class Chrome_WidgetWin_0. Error = 1412
```
I try to fix it now.

CyberTimon:
After the restart, I get a ERR_CONNECTION_REFUSE. Might be related to the vite dev server dying in the background because VS Code thinks the application died. Will investigate furter.

CyberTimon:
Thank you - looks fine to me now and works on Windows / MacOS (tested). I'll merge this. 

Currently we can't change the transparency when running from source, as the vite server / whole npx process dies when the application restarts.

Also currently it still enables transparency by default on linux, because of the new AppDefaults. Do you think this is fine or does this need changes? @andrewazores

andrewazores:
@CyberTimon 
> Also currently it still enables transparency by default on linux, because of the new AppDefaults. Do you think this is fine or does this need changes?

IMO it'd be better to have it off by default on platforms where we know it presents some issues (Linux and Win10), since the user may not realize that there is a setting to disable it, or may not realize that the setting can help with some performance issues. Having the setting and adding documentation about it is good, but setting "safe" defaults for them is even nicer.

PR #50, opened 2025-07-11
ci(asset): set asset name pattern to include platform and arch

### LINKED ISSUE #48: Use Tauri-Actions new filename property to specify version / OS name in the filename
Hey @andrewazores,

Right now, the release is generating two packages for Ubuntu 22.04 and 24.04 — but since the filenames don’t include the version, it ends up being pretty confusing.

About five days ago, this PR (https://github.com/tauri-apps/tauri-action/pull/1099) was merged into the main branch of tauri-action. It’s not in a release yet, but it might be just what we need to fix the issue.

Would you mind taking a look? I can dig into it at some point too, but you already know the CI/CD setup which we use here.

Kind regards,
Timon

andrewazores:
Fixes #48

The CI `build.yml` should be updated at some point to point to a `tauri-action` release version that contains the new feature. For now I have just pointed it directly at the latest commit on the action's dev branch.

Sample runs:
- CI: https://github.com/andrewazores/RapidRAW/actions/runs/16210967639
- Release: https://github.com/andrewazores/RapidRAW/actions/runs/16210974842

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,6 +18,9 @@ on:
       release-id:
         required: false
         type: string
+      asset-name-pattern:
+        required: false
+        type: string
       upload-artifacts:
         required: false
         type: boolean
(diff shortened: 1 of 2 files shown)


CyberTimon:
Thank you! This is amazing and exactly what we want to achieve. I think it's ok when there is still ubuntu in the name. It's a acceptable tradeoff. The ubuntu name can be later.

Will merge this now :)

PR #51, opened 2025-07-11
build(version): simplify release version tagging workflow

andrewazores:
- **set default version to v0.0.0**
- **add tauri-version tool**
- **use release name or commit hash for application version**
- **remove redundant commit hash replacement**

See https://github.com/CyberTimon/RapidRAW/issues/48#issuecomment-3060145294
Separate from #50, but this PR is marked as a draft for now because there will be things to fix up after that one goes in.

Sample runs:
- CI: https://github.com/andrewazores/RapidRAW/actions/runs/16211994088
image
Screenshot_2025-07-11_01-25-17

- Release: https://github.com/andrewazores/RapidRAW/actions/runs/16226626536
image

Regular CI (PRs and merges to `main`) will have assets named like `RapidRAW-0.0.0-3e5638c5643f7acbb215bfbcea2503709e6bbec9-1.x86_64.rpm `. Release assets are named like `RapidRAW-1.2.3-1.x86_64.rpm` (same as now).

These version strings also apply to the actual application itself, so when running an unreleased build, the application actually shows the version string as `0.0.0-$commit_hash`. This way, it's obvious to users or when reading bug reports whether the application version used is actually a release, and if not, the exact commit it was built from.

PR/merge workflow is the same as before, ie do nothing special and the CI system will just produce the builds with the new versioning scheme.

The release workflow is simplified: don't manually update the version strings in the project sources (package.json, tauri.conf.json, Cargo.toml). Just create a new GitHub Release, with a new tag pointing at the desired commit/branch, and name the release like `vX.Y.Z`, creating a git tag by the same name. The CI will take care of patching the project sources within the build environment so that it all matches this release name.


andrewazores:
@CyberTimon ready

CyberTimon:
Just to clarify: After this, the version sync script isn't needed anymore? On release builds, the SHA hash isn't shown in the main welcome screen, right?

Thank you!

andrewazores:
Right, on release builds it's just the release name (semver).

Screenshot_2025-07-11_15-11-20

andrewazores:
And just for completeness, local builds ie `npx tauri dev` or `npx tauri build` will simply end up with `0.0.0`, no commit hash.

CyberTimon:
Thank you for clarifying my questions. I'll merge this soon - LGTM :)

CyberTimon:
I just checked everything and it looks very nice. I really like the way it works. Thank you!

PR #56, opened 2025-07-12
ci(linux): build all pkg formats on all platforms

### LINKED ISSUE #55: AppImage won't run because GLIBC_2.38 is needed
`RapidRAW_v1.2.3_ubuntu_amd64.AppImage` will not launch on my PopOS 22.04 (Ubuntu-based) system because it wants GLIBC_2.38, but my system has GLIBC 2.35.

I think that means RapidRAW needs Ubuntu 24.04. 

Are you able to rebuild the AppImage to use an older version of GLIBC?

andrewazores:
- **only trim '-latest' suffix from build platform, not every last hyphenated portion**
- **build all three pkg formats on all Linux builds**

Fixes #55

Sample runs:
- CI: https://github.com/andrewazores/RapidRAW/actions/runs/16232265170
image

- Release: https://github.com/andrewazores/RapidRAW/actions/runs/16232271116
https://github.com/andrewazores/RapidRAW/releases/tag/v1.1.1
image

@CyberTimon I suggest adding some brief explanation to your Release notes. Probably some users will be confused about aarch64/arm64/amd64/x86_64, or about the Ubuntu/.rpm thing I mentioned elsewhere.


CyberTimon:
Thanks. We're now starting to run into a bit of a problem:

There are so many different Linux builds that MacOS and Windows are getting pushed down. Right now, since the list is sorted alphabetically, the Windows download only shows up under "Show All Assets" - which isn't ideal.

How do you think we should handle this?

andrewazores:
There doesn't seem to be any way to change the sort order, so a couple of options come to mind.

1. Add some other bit of text into the filenames to put the Linux builds at the bottom. For example, renaming the files from "ubuntu" to "z_ubuntu". This might just be confusing.
2. Add links to the release body text pointing users directly to the asset files for MacOS and Windows, since there are only a few options for those, and text telling Linux users to scroll down to find what they need. It might be possible to use the tauri-action releaseBody config to help automatically generate these links? I can test how it behaves when updating a manually created release, and if it overwrites or appends the body or just skips it.

andrewazores:
For 2., the Action won't try to do anything but upload assets if given a releaseId.

If given a tagName instead, then it will create a new release using the releaseName, releaseBody, assets, etc.

https://github.com/tauri-apps/tauri-action/blob/cf3eb9b18add8548a40584695215c80ab7274f31/src/index.ts#L171

So if you're willing to change the release workflow once more, slightly, I can set it up so that the template release body is automatically generated including download links. You would just create a git tag instead of a release. Once the action creates the release you would manually edit it to add whatever other details you want, ie description of what's new.

andrewazores:
Actually, I have an idea to run a separate job after the current one completes, and in that second job use the GitHub API to append to the release body with the generated content. I'll try to implement this option first since it seems like it would be clearest for users and simplest for your workflow.

CyberTimon:
Thank you, that sounds good. I will wait a bit with merging until you have a good conclusion.

andrewazores:
@CyberTimon I see you made some changes to the `ci.yml` file here in https://github.com/CyberTimon/RapidRAW/commit/6166d0ab585a04a8e28a545208e84b012c738fe8 and removed a lot of the CI build types. What was the reason for that? Should I not be working on improvements in this area?

CyberTimon:
Hey @andrewazores,

I trimmed down the CI build types - we were running 7 checks per commit, which feels a bit excessive for simple changes. I’ve left one check per platform, which I think should be sufficient.

Do you see any reason we’d still need to run all 7?

Thanks!

andrewazores:
It depends on what you want the purpose of the CI builds to be. If it's only to check that the application compiles on each platform, then one build for each is enough. But if they are meant to be usable for testing, in particular for community testing of PRs, then having the full list of platforms and architectures is still useful. For example, you dropped ARM/Linux builds from PR/CI, so the user who asked for Asahi Linux support would no longer be able to help test PRs without installing the development toolchain requirements themselves and learning to use GitHub etc. If the ARM/Linux build was still there, we could just point them to the `.deb` file to download and install to check if things are working on their system.

CyberTimon:
You are right - having those builds available does help with community testing. I will (re-)add the missing platforms later.

CyberTimon:
BTW I thought about the naming a bit. What do you think of:

01_windows
02_macos
03_ubuntu
03_ubuntu
03_ubuntu
....

This shouldn't be as confusing as z_ or a_*

What do you think? @andrewazores

andrewazores:
Sure, that's another good option. Maybe that's a good way forward in the short term, while I keep working on trying to add that template message with sectioned download links at some point.

andrewazores:
@CyberTimon ready
1. restores the full os/architecture build matrix to non-release CI
2. builds all Linux pkg formats (deb, rpm, AppImage) on each os/architecture to maximize compatibility (should hopefully fix #55)
3. prefixes Windows builds with `01_`, Mac with `02_`, and Linux with `03_`
4. restores the git commit hash onto the non-release builds' prerelease info. At some point you changed this to be the GitHub Actions run ID number, but that isn't as informative and is harder to actually reference what build someone is running and where it came from. Now this uses `git describe` instead of just the raw commit SHA, so non-release builds look like `03_ubuntu-24.04_RapidRAW-1.1.1-65-gb620f71-1.x86_64.rpm ` - `1.1.1` is the most recent release/tag (on my repo it is anyway) before this non-release build, 65 is the number of commits on this branch since that release, and `gb620f71` is `g` as a prefix meaning `git` with `b620f71` being the hash identifying the particular commit this was built from
5. release builds look like `02_RapidRAW_v1.2.3_macos_aarch64.dmg`

https://github.com/andrewazores/RapidRAW/actions/runs/16306927556

https://github.com/andrewazores/RapidRAW/actions/runs/16306853027

image

Screenshot_2025-07-15_20-08-05

Screenshot_2025-07-15_20-09-37

CyberTimon:
Thank you very much for cleaning this up and implementing the sorting / formats for all plattforms. Looks good to me. 🥇

Issue #76, opened 2025-07-13
A few bugs and Feature Requests


dernikolas:
Bugs:
- At the Export Menu, no matter which format I choose, it always gets exported as .jpg
- I use KDE Plasma and the top bar isn't the default like it should be

<img width="1920" height="35" alt="Image" src="https://github.com/user-attachments/assets/c0cc2985-4d8d-4d68-be23-0e5763eaffde" />

<img width="1920" height="21" alt="Image" src="https://github.com/user-attachments/assets/11a2e204-292c-4435-b167-f06c6852e3a3" />

- In the Library, when I control it with the arrow keys, arrow down doesn't go one down, instead it goes one right (same in reverse with arrow up). Also when I go over the end of the page, it doesn't go down, but the selected image simply goes out of the field of view

Feature Requests:

- Make the Settings reachable from every page, not only the home page
- Make it possible that when I right-click on a photo in the file manager and select "Open with RapidRAW", the photo will immediately open in editing mode (like with Darktable, Lightroom, etc.)

andrewazores:
@CyberTimon this part about "KDE Plasma" is what I mean in #23. A user on KDE, a user on GNOME, a user like myself on i3/Sway, a user on Cinnamon, or XFCE or LXDE or... are all going to have different expectations of what the application titlebar should look like.

dernikolas:
@andrewazores Yes, please implement the native ones

andrewazores:
This is done now since #80 which has been merged, so the latest builds or the next release will have it fixed.

CyberTimon:
> At the Export Menu, no matter which format I choose, it always gets exported as .jpg

I couldn't reproduce this. @andrewazores can you reproduce this on Linux?

andrewazores:
No, I just tested this on latest `main` and I can't reproduce it. I was able to successfully export both PNG and TIFF formats.

I did notice that when I go to the Export Menu, there are the JPEG/PNG/TIFF buttons to select, and then after I confirm my settings and click the "Export Image" button it brings up my OS/DE's file chooser dialog. This dialog has a file type filter, which also presents options for JPEG/PNG/TIFF, but no matter what export format I have selected within RapidRAW's Export Menu, this dialog always defaults to filtering JPEG files. This doesn't actually affect anything though, the file still gets written with the intended format and filename after I select the location I want to save to.

dernikolas:
[...]

PR #80, opened 2025-07-14
feat(linux): use native window decorations

andrewazores:
- **feat(linux): use native window decorations**
- **feat(windows): use transparency effect on windows 11 only**

See #23
See #76


andrewazores:
@CyberTimon round 2 of trying to implement what was in #23. This time there is no user setting, it is just done automatically. This PR should not change anything for MacOS or Windows users. Linux users will get native compositor window decorations (titlebar and controls), so the application will properly match their desktop environment or window manager.

I also tried to handle the Windows 10/11 window transparency bit.

CyberTimon:
Hey!

Thank you a lot for the PR. I will review and merge it in some hours. 

A minor detail I just noticed:
> feat(windows): use transparency effect on windows 11 only

On MacOS we usually also want transparency effects, just not on Linux. 

Will take a deeper look later 💯

andrewazores:
Just an unclear commit message - it only relates to Windows, so only Windows 11 uses transparency and non-11 (ie 10, or 7 or 8 I guess?) does not. MacOS still has transparency enabled and Linux has it disabled.

CyberTimon:
image
On Windows 11 at least, it looks very broken. Top padding is missing and it is just transparent. We want to achieve blur on MacOS / Windows, not transparent / translucent. 

Yes the setting might be a bit misleading..

On MacOS, we want apply_vibrancy and on Windows 11, we want apply_acryl:
image


Are you going to look into this or should I?

Thank you!

andrewazores:
I will look into it later. I only have Linux and one Windows 10 installation so I will check the behaviour on that. I guess I have some logic error there because what I intended was that Windows 10 would have transparency/acrylic disabled (like Linux), but Mac and Windows 11 would have transparency and acrylic/vibrancy enabled, which I think is what you're asking for

andrewazores:
Actually, for now I will just remove the Windows-related changes from this PR and just focus on the Linux native controls aspect. Windows 10/11 acrylic effect handling can be done separately.

andrewazores:
@CyberTimon ready. Tested on Linux and Windows 10. This doesn't touch the transparency or vibrancy/acrylic effects anymore, only the native controls (`.decorations()` function) for Linux builds, and the `App.jsx` css styling that adds that top padding which should be removed when native controls are used.

CyberTimon:
Thanks 💯  - will review & merge this evening.

andrewazores:
@CyberTimon oh whoops, I just got a notification that you added a commit here, at the same time that I rebased and force-pushed it...

CyberTimon:
No problem. I thought I did something wrong :) Lets just add my commit again.

CyberTimon:
Now everything is working as expected on non-linux systems and I think we can merge this - OK? @andrewazores 

Before I had the top padding on the title screen on windows. This isn't optimal so I removed it. I also removed the blur from the titlebar as it doesn't look good.

andrewazores:
Looking good on my Linux system, too.

PR #83, opened 2025-07-14
feat: add nix flake

taciturnaxolotl:
hi! absolutely love this project and wanted to add a nix flake for it :)

you shouldn't need to do anything to maintain it and if it breaks i'll send in a pr to fix it


andrewazores:
I suggest putting it into some subdirectory like `pkg/linux/nix`, so that if/when other similar packaging files might be added they aren't all cluttering the top-level project directory.

CyberTimon:
Thank you @taciturnaxolotl for this PR - that's really cool. I agree with @andrewazores, it's better to keep it in a subdirectory.

I will review it once it's ready 💯

taciturnaxolotl:
i can do that! it's not standard in my experience to have the flake in a sub dir but it is technically supported

andrewazores:
I guess that makes sense for projects that focus specifically on packaging other applications for NixOS, or for projects that are primarily intended for distribution there. It's a bit different nowadays that upstream applications are also so frequently handling their own packaging and distribution efforts - the traditional Linux distribution way of doing things was that the upstream project sources and the downstream package were two completely separate projects, usually even run by different (teams of) people, so you wouldn't really have this kind of concern about different distros' packaging files cluttering up the actual project repository.

taciturnaxolotl:
yeah fair; sticking it in a sub dir is causing a bunch of issues so i'll make my own flake repo for now and then add it to nixpkgs

Issue #75, opened 2025-07-13
Feature Request: Group JPEG and RAW


PixelPusher247:
It would be nice to have an option similar to the Lightroom 'Treat JPEG files next to raw files as separate photos'. In my workflow I capture JPEG and RAW and dump them all in the same folder, afterwards I go through and select/remove pictures, for which having the grouped JPEG and RAW is very convenient when deleting them, so both the JPEG and RAW are deleted.

CyberTimon:
Good idea! I'm considering adding this soon.
Would a third context menu option - "Delete including JPEG" - alongside Delete and Confirm make sense?

PixelPusher247:
Maybe a configuration option to choose if the delete key should delete both?

CyberTimon:
Exactly. I will implement this soon.

CyberTimon:
https://github.com/CyberTimon/RapidRAW/commit/52e3aefe10d73e1b7bb64f6fc5ea7fc122fcf699

<img width="529" height="566" alt="Image" src="https://github.com/user-attachments/assets/ad8d0bdd-26f1-4b93-8861-a9fcd50bcf1f" />

This functionality is added now.

"Grouping" can be achieved by enabling this filter:

<img width="409" height="182" alt="Image" src="https://github.com/user-attachments/assets/7b34302b-50b5-42cb-b867-c7decae2e6b5" />

PR #89, opened 2025-07-15
feat: Add filter options for file types; is_edited

ALPHAAAL:
Just thought that this is a nice filter to have since sometimes when importing files it contains both JPEG and raws.

### DIFF
--- a/src-tauri/src/file_management.rs
+++ b/src-tauri/src/file_management.rs
@@ -27,6 +27,20 @@ const THUMBNAIL_WIDTH: u32 = 640;
 pub struct ImageFile {
     path: String,
     modified: u64,
+    is_edited: bool,
+}
+
+fn has_sidecar_adjustments(image_path: &str) -> bool {
+    let sidecar_path = get_sidecar_path(image_path);
+    
+    // Use file size as a fast heuristic:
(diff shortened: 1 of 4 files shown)


ALPHAAAL:
Found out this issue, kind of related
https://github.com/CyberTimon/RapidRAW/issues/75

CyberTimon:
Thanks for the PR, @ALPHAAAL! These are very useful filter additions 💯 

I've noticed a technical issue with the 'edited/unedited' filter. The is_edited status is only determined when the folder is first loaded and doesn't update when a user makes or resets edits. This causes the filter to become inaccurate.

For that reason, I'll be removing the edited/unedited filter for now. The RAW / Non-RAW filtering is a solid addition, and I'll merge that part of the change.

If you can find a good solution for updating the is_edited state accurate, I'd be happy to review and merge that functionality as well.

P.S. I also updated the has_sidecar_adjustments check in the backend. I'm leaving this improved functionality in place as it will be useful for future filtering.

Thanks again for the contribution :)

ALPHAAAL:
Hey @CyberTimon, thanks for the review and you have a great point! Let me get back to this later 👍

william-sy:
Related issue: https://github.com/CyberTimon/RapidRAW/issues/190

Issue #38, opened 2025-07-09
Sluggish window moving


fcmss:
Hello!
It's hard to describe the issue (mainly because of language barrier) but the main window becomes very sluggish when you move it (at least on my Windows 10 machine).
I don't know how much it shows on the video recording, but it's VERY sluggish. I feel like there's 1 or 2 seconds of delay between when I release the window and when it actually let my cursor comes back. This happens on every part of the app (not just the landing page).
There's maybe some transparency effect or something that takes a lot of power?
I don't know what to give you in order to try and debug this?

My specs:
Windows 10 Pro Build 19045
i7-9750H
RTX 2060 (on a laptop)

Running RapidRAW 1.2.1.

https://github.com/user-attachments/assets/b63e54dc-4059-4a2a-8ea9-5ddb368df1d5

CyberTimon:
Thank you for the report. It's a duplicate of https://github.com/CyberTimon/RapidRAW/issues/13.

It's a known Windows 10 bug and this also happens on some other applications.

References:
https://github.com/seo-rii/electron-acrylic-window/issues/6
https://github.com/tauri-apps/window-vibrancy/issues/47
https://www.google.com/search?q=windows+acrylic+lag

It looks like it might be related to the mouse pooling rate. I might just provide an option in the near future to disable the blur (in the RapidRAW settings).

Kind regards,
Timon

fcmss:
> Thank you for the report. It's a duplicate of [#13](https://github.com/CyberTimon/RapidRAW/issues/13).
> 
> It's a known Windows 10 bug and this also happens on some other applications.
> 
> References: [seo-rii/electron-acrylic-window#6](https://github.com/seo-rii/electron-acrylic-window/issues/6) [tauri-apps/window-vibrancy#47](https://github.com/tauri-apps/window-vibrancy/issues/47) https://www.google.com/search?q=windows+acrylic+lag
> 
> It looks like it might be related to the mouse pooling rate. I might just provide an option in the near future to disable the blur (in the RapidRAW settings).
> 
> Kind regards, Timon

Oh, my bad, I looked at previous issues but couldn't find one (seems like I can't read).
Thanks for the comment.

What would be the best option? Disabling vibrancy if system is Windows 10? Disabling vibrancy when resizing/moving?
The problem I see with hiding it in the options is that people will assume it is the software's fault and that there's nothing they can do to fix that.
Kind regards,
fcmss

andrewazores:
[...]

PR #92, opened 2025-07-15
fix(windows): use apply_acrylic on Windows 11+ and apply_blur on Windows <= 10

andrewazores:
See #80
See #38


andrewazores:
@CyberTimon tested on Windows 10 and this fixes the sluggish window moving issue for me.

CyberTimon:
Amazing! Big thanks for looking into this bug. Tell me when you are ready with this :) 🥇

andrewazores:
It's also looking good on Linux (no apparent changes, as expected). If you can test that it looks the same as usual on Win11/MacOS then this is ready.

andrewazores:
Also, I didn't bother with trying to cache/persist the detected OS information as per https://github.com/CyberTimon/RapidRAW/issues/38#issuecomment-3060857307 . I haven't profiled it myself but if we assume that that other commenter's figures are roughly accurate still, 35ms is not a lot of additional startup time. If it becomes a problem then some logic to persist the information to ex. the app settings file could be done, but then this runs into questions about what happens if the user upgrades OS versions, for example.

CyberTimon:
> Also, I didn't bother with trying to cache/persist the detected OS information as per https://github.com/CyberTimon/RapidRAW/issues/38#issuecomment-3060857307 . I haven't profiled it myself but if we assume that that other commenter's figures are roughly accurate still, 35ms is not a lot of additional startup time. If it becomes a problem then some logic to persist the information to ex. the app settings file could be done, but then this runs into questions about what happens if the user upgrades OS versions, for example.

Yes I agree this shouldn't be a big problem and only affects Linux.

I just tested it on Windows 11 and MacOS and on both systems, the transparency / blur / acryl / whatever is completely gone -> completely black container. On Windows 11, when I move the window with the titlebar, the transparency glitches through (it's translucent in some frames, while moving).

CyberTimon:
> [re-add maybe not unused dependency?](https://github.com/CyberTimon/RapidRAW/pull/92/commits/179a1c433a980604c595fb3319d4245da2e34f8b)

Oh yes - I might have to tell you that these dependencies get reported as unused by rust-analyzer. I will recheck soon.

CyberTimon:
https://github.com/user-attachments/assets/e83fbda1-277d-4c3e-b990-693af531efba

The issue is sadly still there. First RapidRAW instance is your PR and the next RapidRAW instance is v1.2.5. Also notice how the greyish black is almost complete black on your PR.
I can take a look at this soon if you want / or don't have the capabilities to debug this further, as it's Windows 10 / 11 related.

andrewazores:
@CyberTimon sure, please continue on with this work then. I won't be able to take it much further since it is platform-specific.

CyberTimon:
Will do - I think os_info is buggy

CyberTimon:
[Window Effect] Full OS Info: Windows 10.0.26100 (Windows 11 Professional) [64-bit]
[Window Effect] Detected Version: Semantic(10, 0, 26100)
[Window Effect] Is this Windows 11 or newer? -> false
[Window Effect] Applying Blur effect for Windows 10 or older.

great it detects Windows 11 as Windows 10 😆

andrewazores:
Nice... well it pulls that "Windows 11 Professional" string out and IIRC when I was looking at it, the `Version` struct can be formatted as a string, so I guess you could check if the string contains `"Windows 11"`. But that's far from ideal and is pretty fragile.

I know you weren't a fan of the approach before, but I suggest considering a settings menu item to enable/disable transparency and effects.

andrewazores:
Actually, that doesn't seem to be a bug in os_info, that's actually just how Microsoft decided to tag these Windows versions:

https://en.wikipedia.org/wiki/Windows_11_version_history#Version_history

CyberTimon:
Yes I also just saw this. Just doing a string check for "Windows 11" would fail on most non-English systems.

CyberTimon:
I'm gonna merge this now as I haven't changed anything on Linux / MacOS and it works correctly on Windows 11 / 10.

Issue #82, opened 2025-07-14
System requirements


hp48gx:
Thanks for your excellent work (from a photographer).

Can you please list somewhere the system requirements? For example, on my old Max OS laptop, it crashes on startup (see the report below).
Similarly, I don't see any info for Windows: I guess from the open issues that it may have problems on Windows 11, but will it work -say- on Windows 7?

```
Process:               RapidRAW [8136]
Path:                  /Applications/RapidRAW.app/Contents/MacOS/RapidRAW
Identifier:            com.rapidraw.app
Version:               1.2.5 (1.2.5)
Code Type:             X86-64 (Native)
Parent Process:        ??? [1]
Responsible:           RapidRAW [8136]
User ID:               501

Date/Time:             2025-07-14 21:50:08.945 +0200
OS Version:            Mac OS X 10.14.6 (18G9323)
Report Version:        12

System Integrity Protection: enabled

Crashed Thread:        0

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Reason:    DYLD, [0x4] Symbol missing

Application Specific Information:
dyld: launch, loading dependent libraries

Dyld Error Message:
  Symbol not found: _NSHTTPCookieSameSiteLax
  Referenced from: /Applications/RapidRAW.app/Contents/MacOS/RapidRAW
  Expected in: /System/Library/Frameworks/Foundation.framework/Versions/C/Foundation
 in /Applications/RapidRAW.app/Contents/MacOS/RapidRAW
```

hp48gx:
(personal opinion here: if you disagree just ignore me...)

I know a lot of photographers who are still on Windows7, precisely because they installed Lightroom6 with a perpetual license... so being win7-compatible would be a _very_ strong plus.
Otherwise none of these photographers will really consider evaluating RapidRaw.

andrewazores:
> but will it work -say- on Windows 7?

The fastest way to get an answer to that question is probably to just ask one of those photographers you know to test it :-)

hp48gx:
That I may check. Was the failure on macos 10.14 expected?

andrewazores:
@CyberTimon any idea about that? GitHub has runners for MacOS 13 and 14 as well, maybe the CI should build on one of those instead of latest/15, or on all of them?

CyberTimon:
@andrewazores 

[...]

PR #102, opened 2025-07-16
build(macos): set minimum system version 10.13

andrewazores:
See #82


andrewazores:
@CyberTimon it seems like 10.13 will not actually work (refer back to discussion in#82), even if this property is set in the macOS bundle configuration. Should I set this instead to 15.0, or some other value like that, so that users on older macOS versions should see some kind of incompatibility notice? I see that in the project README you specified macOS 15+ is required - is that for sure the case, or could some older versions also work?

CyberTimon:
> I see that in the project README you specified macOS 15+ is required - is that for sure the case, or could some older versions also work?

It's the last known version which works and doesn't have this missing symbol error. Building RapidRAW on older runners might be the solution.

> Should I set this instead to 15.0, or some other value like that, so that users on older macOS versions should see some kind of incompatibility notice?

Until we find a better solution, this might be a good fix.

Thank you!

andrewazores:
@CyberTimon sounds good. Ready for review then. I'll open a separate PR to experiment with lowering the minimum version, but for now this one just sets it to 15 to match the README documentation.

CyberTimon:
Thank you. Looks good to me 👍

CyberTimon:
MacOS runners aren't working anymore:
https://github.com/CyberTimon/RapidRAW/actions/runs/16526002186/job/46739349886

I can take a look at this soon, or do you know a fix already? @andrewazores

andrewazores:
Weird that it worked in PR CI but not after merge. Just removing the "frameworks" part of the conf that I added should be fine. I think this Foundation framework is already implicitly included anyway.

CyberTimon:
Perfect. Will try this with the next commit.

PR #147, opened 2025-07-25
feat: add categories for desktop file

taciturnaxolotl:
hi! i'm currently packaging rapidraw for nix via nixpkgs and I noticed that the tauri generated desktop file doesn't have proper categories on linux so I thought I'd add them :)

### DIFF
--- a/src-tauri/tauri.linux.conf.json
+++ b/src-tauri/tauri.linux.conf.json
@@ -15,7 +15,9 @@
   "bundle": {
     "linux": {
       "rpm": {
-        "provides": ["rapid-raw"],
+        "provides": [
+          "rapid-raw"
+        ],
         "recommends": [
           "webkit2gtk4.1",
           "webkit2gtk4",
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -21,12 +21,15 @@
     ],
     "security": {
       "csp": null,
-      "capabilities": ["default"]
+      "capabilities": [
+        "default"
+      ]
     },
     "macOSPrivateApi": true
   },
   "bundle": {
     "active": true,
+    "category": "Photography",
     "targets": "all",
     "icon": [
       "icons/32x32.png",


src-tauri/tauri.linux.conf.json
```diff
@@ -14,8 +14,15 @@
   },
   "bundle": {
     "linux": {
+      "categories": [
```
andrewazores, line 17: Is there any documentation about this config property? I don't see it listed in the Tauri config schema: https://schema.tauri.app/config/2

andrewazores:
I unpacked the 24.04 x86_64 .rpm produced by CI for this PR, and the desktop file within looks like this:

```
[Desktop Entry]
Categories=
Comment=A blazingly-fast, non-destructive, and GPU-accelerated RAW image editor built with performance in mind.
Exec=RapidRAW
Icon=RapidRAW
Name=RapidRAW
Terminal=false
Type=Application
```

The Categories are still empty, so it seems like Tauri is ignoring this unrecognized config property.

taciturnaxolotl:
yeah i got the property wrong; as per the [schema](https://schema.tauri.app/config/2) it should be

```json
"category": {
  "description": "The application kind.\n\n Should be one of the following:\n Business, DeveloperTool, Education, Entertainment, Finance, Game, ActionGame, AdventureGame, ArcadeGame, BoardGame, CardGame, CasinoGame, DiceGame, EducationalGame, FamilyGame, KidsGame, MusicGame, PuzzleGame, RacingGame, RolePlayingGame, SimulationGame, SportsGame, StrategyGame, TriviaGame, WordGame, GraphicsAndDesign, HealthcareAndFitness, Lifestyle, Medical, Music, News, Photography, Productivity, Reference, SocialNetworking, Sports, Travel, Utility, Video, Weather.",
  "type": [
    "string",
    "null"
  ]
},
```

src-tauri/tauri.linux.conf.json
```diff
@@ -14,8 +14,11 @@
   },
   "bundle": {
     "linux": {
+      "category": "Photography",
```
andrewazores, line 17: I think this should be directly under `bundle`, not within `linux`, if I'm reading the schema correctly.

taciturnaxolotl: oh yeah

taciturnaxolotl:
hmm that didn't do it either

image

andrewazores:
This still doesn't seem to have set the expected category :-(

```
$ cat usr/share/applications/RapidRAW.desktop 
[Desktop Entry]
Categories=
Comment=A blazingly-fast, non-destructive, and GPU-accelerated RAW image editor built with performance in mind.
Exec=RapidRAW
Icon=RapidRAW
Name=RapidRAW
Terminal=false
Type=Application
```

andrewazores:
It seems the nixpkg PR was already accepted even without this being fixed, but @taciturnaxolotl if you're still looking to fix this, I think this might be the approach needed:

https://github.com/tauri-apps/tauri/pull/5180

taciturnaxolotl:
Ooh thanks! I guess i was just looking in the wrong place lol

andrewazores:
I'm not really sure. From the looks of that old PR, there should be a default desktop template which includes some categories from the config, but this doesn't seem to actually be working. I haven't tried to dig deeper into why not. Setting a desktop template directly might be a reasonable workaround for that apparent default one not rendering the categories as expected.

taciturnaxolotl:
i'm just going to fix it manually in the package

PR #148, opened 2025-07-25
build(macos): lower minimum version to 13.0

andrewazores:
Depends on #102 

Sets the minimum version in the application bundle, and the GitHub Actions runner used to build macOS distributions, to 13 (Ventura).


andrewazores:
@CyberTimon up to you if you'd like to accept this, since I have no way of testing on macOS 13/14 and it doesn't sound like you do either :-)

CyberTimon:
Thank you @andrewazores. I will merge this when I‘m home. If it doesnt break the existing macOS 15 support, then I think we‘re all good.

PR #157, opened 2025-07-28
feat(macos): green titlebar button fullscreens window

### LINKED ISSUE #156: True fullscreen mode on MacOS
First of all I really like RapidRAW as it's the easiest open source raw editor to start with. I now recommend it to people who start out with photography instead of darktable.
Nonetheless, I noticed on my MacBook, I cannot enter the complete fullscreen mode with neither 🌐+f nor the green button at the upper left of the window.
The window maximizes, but I want the view without the menu bar and the dock. For comparison, I'll add Lightroom screenshots.

Image

Image

This may be a small thing, but usefull with smaller screens.

andrewazores:
Fixes #156

Not sure how to bind this Mac "globe key", or if it's even possible to do so with Tauri. I'm looking at https://v2.tauri.app/plugin/global-shortcut/ but I don't see any modifier like "globe" as an option and I don't know what else it might be aliased as, if anything.

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -3,6 +3,7 @@ import { invoke } from '@tauri-apps/api/core';
 import { listen } from '@tauri-apps/api/event';
 import { open } from '@tauri-apps/plugin-dialog';
 import { homeDir } from '@tauri-apps/api/path';
+import { getCurrentWindow } from '@tauri-apps/api/window';
 import debounce from 'lodash.debounce';
 import { centerCrop, makeAspectCrop } from 'react-image-crop';
 import clsx from 'clsx';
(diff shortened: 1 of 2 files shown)


andrewazores:
@skranz0 @CyberTimon please let me know if this behaves as expected on macOS

CyberTimon:
When I start the application on macOS, I get a React is not defined error. After adding it, I get a 

> TypeError: undefined is not an object (evaluating 'React.useCallback')

error. Should I try to fix some stuff or will you do?

andrewazores:
@CyberTimon right, sorry, too used to how we do React imports on my other project. And no TypeScript (yet!) here to catch me out at build time :-/

CyberTimon:
Now the titlebar also hides automatically when it's fullscreen. MacOS shows the native titlebar when moving the mouse to the the top of the screen, so this would have cause 2 titlebars.

CyberTimon:
I merge this, ok? @andrewazores

andrewazores:
@CyberTimon good to go!

PR #161, opened 2025-07-28
remove src-tauri/resources shared libraries from git tracking

### LINKED ISSUE #159: Linux ARM CI builds fail on `linuxdeploy`
https://github.com/CyberTimon/RapidRAW/actions/runs/16575698467/job/46879333119

```
running npm [ 'run', 'tauri', 'build' ]
> rapidraw@1.3.1-7-g7020d45 tauri
> tauri build
     Running beforeBuildCommand `npm run build`
> rapidraw@1.3.1-7-g7020d45 build
> vite build
vite v6.3.5 building for production...
transforming...
✓ 2283 modules transformed.
rendering chunks...
computing gzip size...
dist/index.html                   0.81 kB │ gzip:   0.44 kB
dist/assets/index-POcPOuOl.css   41.60 kB │ gzip:   7.43 kB
dist/assets/index-BQkSX07L.js   952.03 kB │ gzip: 286.78 kB

(!) Some chunks are larger than 500 kB after minification. Consider:
- Using dynamic import() to code-split the application
- Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/configuration-options/#output-manualchunks
- Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
✓ built in 4.73s
    Updating crates.io index
   Compiling rawler v0.7.0 (/home/runner/work/RapidRAW/RapidRAW/src-tauri/rawler)
   Compiling RapidRAW v1.3.1-7-g7020d45 (/home/runner/work/RapidRAW/RapidRAW/src-tauri)
warning: RapidRAW@1.3.1-7-g7020d45: Could not find ONNX Runtime library at: /home/runner/work/RapidRAW/RapidRAW/src-tauri/target/release/libonnxruntime.so. It might be copied on a subsequent build.
warning: unused imports: `NSVisualEffectMaterial`, `apply_acrylic`, and `apply_vibrancy`
  --> src/main.rs:30:23
   |
30 | use window_vibrancy::{apply_acrylic, apply_vibrancy, NSVisualEffectMaterial};
   |                       ^^^^^^^^^^^^^  ^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
warning: unused variable: `theme`
   --> src/main.rs:995:24
    |
995 | fn apply_window_effect(theme: String, window: impl raw_window_handle::HasWindowHandle) {
    |                        ^^^^^ help: if this is intentional, prefix it with an underscore: `_theme`
    |
    = note: `#[warn(unused_variables)]` on by default
warning: unused variable: `window`
   --> src/main.rs:995:39
    |
995 | fn apply_window_effect(theme: String, window: impl raw_window_handle::HasWindowHandle) {
    |                                       ^^^^^^ help: if this is intentional, prefix it with an underscore: `_window`
warning: `RapidRAW` (bin "RapidRAW") generated 3 warnings (run `cargo fix --bin "RapidRAW"` to apply 1 suggestion)
    Finished `release` profile [optimized] target(s) in 6m 11s
       Built application at: /home/runner/work/RapidRAW/RapidRAW/src-tauri/target/release/RapidRAW
    Bundling RapidRAW_1.3.1-7-g7020d45_arm64.deb (/home/runner/work/RapidRAW/RapidRAW/src-tauri/target/release/bundle/deb/RapidRAW_1.3.1-7-g7020d45_arm64.deb)
    Bundling RapidRAW-1.3.1-7-g7020d45-1.aarch64.rpm (/home/runner/work/RapidRAW/RapidRAW/src-tauri/target/release/bundle/rpm/RapidRAW-1.3.1-7-g7020d45-1.aarch64.rpm)
 Downloading https://github.com/AppImage/AppImageKit/releases/download/continuous/AppRun-aarch64
 Downloading https://github.com/tauri-apps/binary-releases/releases/download/linuxdeploy/linuxdeploy-aarch64.AppImage
 Downloading https://raw.githubusercontent.com/tauri-apps/linuxdeploy-plugin-gtk/master/linuxdeploy-plugin-gtk.sh
 Downloading https://raw.githubusercontent.com/tauri-apps/linuxdeploy-plugin-gstreamer/master/linuxdeploy-plugin-gstreamer.sh
    Bundling RapidRAW_1.3.1-7-g7020d45_aarch64.AppImage (/home/runner/work/RapidRAW/RapidRAW/src-tauri/target/release/bundle/appimage/RapidRAW_1.3.1-7-g7020d45_aarch64.AppImage)
failed to bundle project: `failed to run linuxdeploy`
       Error failed to bundle project: `failed to run linuxdeploy`
```

https://github.com/tauri-apps/tauri/issues/13113

https://github.com/tauri-apps/tauri/issues/8929

This only seems to affect the Linux (Ubuntu) builds on ARM architecture, and somehow only started to happen after #158.

The `.so` that I added in #158 came from my x86_64 system, so it would be the wrong one for an ARM system. However, based on the build log output, it doesn't look like the build is picking up the presence of this file anyway, so I'm not sure if that's actually the problem.

andrewazores:
Fixes #159


andrewazores:
@CyberTimon when I do `npx tauri dev` I see `src-tauri/target/debug/` contains `libonnxruntime.so` and `libonnxruntime.so.1.16.0`, which appear to be identical copies. Likewise `npx tauri build` produces `src-tauri/target/release` which contains the same two `.so` files. Where do these come from? I thought it would just be copying the contents of `src-tauri/resources`, but that doesn't seem to be what actually happens.

CyberTimon:
Hey @andrewazores 

They get downloaded automatically by the onnxruntime ORT crate: https://docs.rs/ort/latest/ort/#how-to-get-binaries

Hope this helps.

andrewazores:
Nice, I had a feeling - then in that case, what are the `src-tauri/resources` copies used for? It seems like we should only need one or the other, no?

CyberTimon:
the build script tries to find the automatically downloaded binaries form these folders and then copies them to the src-tauri/resources directory, which then gets packed as defined in the tauri config:
```
    "resources": [
      "resources",
      "workflows"
    ],
```

CyberTimon:
> Nice, I had a feeling - then in that case, what are the `src-tauri/resources` copies used for? It seems like we should only need one or the other, no?

Yes I also don't know why the ORT crate downloads 2 (one with version number and one without).

andrewazores:
@CyberTimon ready for review. I tested this with local `npx tauri dev` and `.rpm` builds, and AI features like foreground and subject selection are working as expected (Linux x64).

1. prevents `ort` crate from redownloading runtime dynamic libraries, so that the ones checked-in to git within the project are always used
2. updates the `build.rs` file path to append `resources` to the lookup path, so that the checked-in runtime library is used
3. renames `libonnxruntime.so` to `libonnxruntime.x64.so`
4. adds `libonnxruntime.aarch64.so`, which I downloaded from https://github.com/microsoft/onnxruntime release 1.16.0, since that seems to match the version used previously
5. adds compile-time checks so that the Linux builds attempt to load the appropriate dynamic library filename for the build architecture
6. sets an environment variable in CI so that the Linux builds should not be "confused" by the fact that `.so` files for a "wrong" build architecture are present (since both versions are always present, but have different names and only the correct one should be loaded at runtime). Since this is within `.github/workflows` and the CI runs on `pull_request_target` event, this change is not effective within this PR and won't apply until after the PR is merged to `main`, so Linux PR CI should be expected to fail here.
7. in `build.rs`, when checking if the runtime file is up-to-date, compare SHA256 file hashes rather than the entire file byte contents. This should just make building maybe a little faster, and take a tiny bit less memory

CyberTimon:
Great! Thank you a lot for optimizing / cleaning this up. Do you know why the windows build fails now? I can try to build it fresh on my windows / macos machine soon, after I finished some of the ai generative features refactoring.

andrewazores:
I'm not sure, I'm re-checking my Linux build and watching the CI output to make sure it isn't something obvious and cross-platform. If it is Windows-specific then I really don't know.

CyberTimon:
Also, doesn't the same issue also arrise now on macos arm / x64?

Edit: Maybe it's windows rust build cache related?

andrewazores:
> Also, doesn't the same issue also arrise now on macos arm / x64?

I would think so, yes. I can repeat the same fixes here for macOS easily enough, but (as usual) can't test the results.

There is also the option of creating "universal" macOS builds, and using the corresponding universal onnxruntime shared library, but as far as I can tell this basically just means that the binaries for both architectures are bundled together so that the system can load the relevant one - so everything would be about 2x the size.

andrewazores:
Local fresh rebuild on Linux succeeded and is working fine, and CI successfully built a macOS/ARM run, too, so the Windows thing does look like it's just a cache issue or something similar.

andrewazores:
macOS/Intel build has also succeeded. But, I assume that the `libonnxruntime.dylib` being used will now be wrong for one of those builds, since it should be the one that is checked in to the project. So despite the build passing the AI features may not actually work. I guess that the `.dylib` is an ARM one, so the Intel build would probably fail to load it.

CyberTimon:
So macOS Intel build's AI features also weren't working before, right? I will test this on windows now

CyberTimon:
Unfortunately, it also fails on Windows - I’m unable to build due to a linking issue related to Advapi32. Are you sure this is the best approach, particularly with not downloading the libraries on the fly during the build process? It seems like this method is introducing more issues than it resolves. (Please don’t take this as a criticism of your work)

EDIT: Even on a fresh cargo clean build

andrewazores:
> So macOS Intel build's AI features also weren't working before, right? I will test this on windows now

I'm not sure, actually. I think the fact that the `ort` crate was set up to try to dynamically download a runtime shared library file could have been keeping it working. Without the `build.rs` change to append `"resources"` to the path, I think `src-tauri/resources` copies were actually being ignored this whole time and `ort`'s downloaded runtimes were being used instead. IMO it makes enough sense for models to be downloaded on demand, but the runtime itself should probably be bundled with RapidRAW.

But, if we (you) think it makes sense to allow `ort` to download the runtime on demand, then I think actually we can skip this PR and do almost the opposite instead: remove `src-tauri/resources` completely so that no onnxruntime files are distributed as part of RapidRAW, and just let it get downloaded on demand. Presumably the `ort` crate will handle downloading the correct one for the OS/architecture. I assume you had some reason to initially keep the Windows and macOS shared libraries in `src-tauri/resources`, though.

CyberTimon:
> remove src-tauri/resources completely so that no onnxruntime files are distributed as part of RapidRAW, and just let it get downloaded on demand

we still need this, as it doesn't download the onnxruntime on the fly while starting the application, instead it downloads them on the fly, while building. (and if it isn't included in the build, it fails while loading)

> I assume you had some reason to initially keep the Windows and macOS shared libraries in src-tauri/resources, though.

To be honest I just forgot to add them to the .gitignore, that's why they started to appear in the repository.

> IMO it makes enough sense for models to be downloaded on demand, but the runtime itself should probably be bundled with RapidRAW.

Yes definitely. Models should be downloaded on demand (like it's implmeneted currently and it has nothing to do with the binaries issue we're having here). The binaries should be bundled with RapidRAW, preferably in the src-tauri/ressources directory, as it's known to work when they are placed there.

andrewazores:
> we still need this, as it doesn't download the onnxruntime on the fly while starting the application, instead it downloads them on the fly, while building. (and if it isn't included in the build, it fails while loading)

I'm not sure if this makes sense, though. If `ort` already downloads the library files, which I saw it placing into `src-tauri/target/debug/` and `src-tauri/target/release`, that should be already enough - why keep the `src-tauri/resources` copies? What should they be used for?

I think it should be possible to just have `ort` download the files, include those in the build, and have RapidRAW load those when it starts.

andrewazores:
Wait, I think we're just miscommunicating. I just mean removing `src-tauri/resources` __from git__, so they're no longer part of the RapidRaw _project_. The libraries would still need to be included in the distributed RapidRAW packages of course, which I think is what you mean :-)

CyberTimon:
> why keep the src-tauri/resources copies? What should they be used for?

While integrating ORT, tauri didn't package the binaries into the final bundle (when they are just laying around in release/debug targets and thus main.rs couldn't find / load them.

CyberTimon:
Yes exactly :-) We're definitely miscommunicating. I had a hard day and I need to go to sleep 😆. I will take a look at here tomorrow, probably after work. Thank you again!

andrewazores:
No worries, I didn't catch on either.

So then in that case, I will rework this PR (or abandon it and open a new one maybe) that does that - removes `src-tauri/resources` from git, and just lets `ort` download the correct dependency at compile time. I think this should also fix #159 since then there should no longer be any cross-architecture shared library files that confuse the appimagetool.

CyberTimon:
Amazing! Are you ready with this? Have you checked if onnxruntime still gets included (it definitely should but we never know)

andrewazores:
Yes, ready. I did a local build and verified that the .so file is included as expected, and that AI subject detection still works. I also checked and the .dll and .dylib are no longer included, so the filesize got notably smaller. I haven't checked the Windows or macOS builds but I see no reason that it would be any different.

CyberTimon:
Thank you so much. Looks good to me 💯

PR #166, opened 2025-07-29
Draft: Refactor UI to TypeScript

varjolintu:
Refactors the whole UI to TypeScript. Adds ESLint and Prettier for code formatting.

PR #188, opened 2025-07-31
LUT Implementation Phase 1

GoldenWRaft:
# LUT Implementation Phase 1

@CyberTimon here are some changes for the LUT implementation to this point, I'll continue with the implementation but the state and image settings I would probably need help with implenting


andrewazores:
I suggest @CyberTimon creates a new feature branch on the repo and this PR is targeted to merge into that branch, rather than directly into `main`. Once you two have everything implemented as you'd like and the feature is complete, then that feature branch can be merged into `main` all at once. This way things can keep moving and it's easier to collaborate, without the work in progress ending up directly in `main` and half-finished work either blocking releases or ending up released unfinished.

CyberTimon:
Thanks, @andrewazores, for the tip - we should definitely implement this! 👍 I'm no Git expert, so these kinds of suggestions are always appreciated.

And @GoldenWRaft, huge thanks for implementing LUTs support! 🙌 I don’t have much experience with LUTs, but I can already see the value in having this. Let me know when you're ready - I will take a look then! 💯

GoldenWRaft:
@CyberTimon you can create a a feature branch that I can do a pull request into there and then you can see what I am up to. The settings and the flow is a bit foreign to me at this point but we can collaborate on this one, I did not think it will be such a Massive implementation. But now I understand it touches the core of this application

CyberTimon:
Will do that soon. Btw @GoldenWRaft how is the performance? The processing probably doesn't happen in a single GPU pass anymore, right? Does this work on very large images, e.g. 10000x6000 pixels?

GoldenWRaft:
> Will do that soon. Btw @GoldenWRaft how is the performance? The processing probably doesn't happen in a single GPU pass anymore, right? Does this work on very large images, e.g. 10000x6000 pixels?

The performance is quite good, I think, but it should be about the same to the shader.wgsl file that already ran. I do think it works with the 10k x 6k pixels files

GoldenWRaft:
@CyberTimon 

I'm not sure if I showed this off but this is how the process looks like at this point:
image

You are able to add multiple folder locations across different locations and will be a list of items like this:
image

The idea will be to save all of these to the config file and will store it across sessions

CyberTimon:
Thank you. I will definitely take a look on the weekend. Had lots to do this week with work and stuff 👍

CyberTimon:
Hey @GoldenWRaft 

In the meantime, we merged https://github.com/CyberTimon/RapidRAW/pull/166, which refactors the UI to TypeScript. Small adjustments and a rebase is needed for this PR. If you have any questions, feel free to ping @varjolintu or me.

Thank you!

GoldenWRaft:
Cool, I will do that and notify when it is ready

GoldenWRaft:
@CyberTimon I naively converted my implementation to typescript. I have not added predefined types but instead marked as inline objects.

CyberTimon:
@GoldenWRaft Hey, I just tested it out and, to be honest, it feels a bit buggy and unstable right now. It doesn’t seem to save the LUT in the adjustments payload or sidecar, so the LUT gets lost whenever you make further adjustments. Also, the thumbnails don’t update correctly - this might be related to the missing save functionality. Another thing: when I open the LUTs panel and switch to a different panel, it resets its state.

By the way, which AI did you use for this implementation? I noticed some GPT-4o like comments. From my experience, using GPT-4o for RapidRAW tends to produce lower-quality code, which is why I’ve always preferred Google AI Studio - it consistently delivers much better results. No offense intended!

If you prefer, I can make the frontend and we somehow merge the backend code?

GoldenWRaft:
Yes, I used GPT 4 for the frontend changes, and the saving the settings is a bit tricky for me as I do not know the entire flow. The rest of the implementation should be mostly frontend. I also used Google's AI studio in combination to do the backend and frontend

GoldenWRaft:
> If you prefer, I can make the frontend and we somehow merge the backend code?

@CyberTimon Yes please, that will work for me. Otherwise how is the performance and the changes that I've made?

CyberTimon:
The performance for applying single LUT's is definitely OK :). I try to make some progress this weekend regarding LUT's and I will definitely take some of your backend code. I'm still wondering if it's better to apply the LUT in the same wgsl shader as this would keep everything in a single GPU pass.

GoldenWRaft:
@CyberTimon it is worth a test though. Depending on what type of system the user is running it might improve the performance or slightly degrade. Unfortunately I'm no expert

CyberTimon:
Hello @GoldenWRaft,

I’ve just finished implementing LUT support from the ground up. Doing everything in a single GPU pass fits much better with the rest of the RapidRAW core, and having proper saving & caching of the loaded LUTs is also an important part of the workflow.

That said, I still want to thank you for your contribution. Even though I won’t be merging it, I really appreciate the effort you put in - you definitely gained experience along the way!

Commit: https://github.com/CyberTimon/RapidRAW/commit/47c2561b2c99e08effb163306cae51cec4549cc9

Kind regards,
Timon

GoldenWRaft:
@CyberTimon Thank you for looking at my work, please continue with this great app. I will definitely use this app for all my image processing.

PR #203, opened 2025-08-03
feat: add edited indication on main gallery

ipetrovbg:
# This PR

- Adds an icon indication when an image is edited to help visually identify processed images

Main Gallery
|Before|After|
|---|---|
|image|image|

Film Strip
|Before|After|
|---|---|
|image|image|

### DIFF
--- a/src/components/panel/Filmstrip.jsx
+++ b/src/components/panel/Filmstrip.jsx
@@ -1,5 +1,5 @@
 import { useEffect, useRef } from 'react';
-import { Image as ImageIcon, Star } from 'lucide-react';
+import { Image as ImageIcon, Star, Pencil } from 'lucide-react';
 import { motion, AnimatePresence } from 'framer-motion';
 
 const COLOR_LABELS = [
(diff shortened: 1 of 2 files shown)


CyberTimon:
Thank you for the PR! 

How does it handle it when you edit a image and then jump back into the library, or reset the adjustments on a image? Does it correctly update the thumbnail edited state or does it become stale?

We had a similar problem here:

https://github.com/CyberTimon/RapidRAW/pull/89#issuecomment-3074316404

ipetrovbg:
Yeah, it doesn't really. It uses the state on the image `is_edited` which I suppose (without looking into it in dept) is probably using the modified file descriptor.

ipetrovbg:
> Yeah, it doesn't really. It uses the state on the image `is_edited` which I suppose (without looking into it in dept) is probably using the modified file descriptor.

Actually that's not true. It uses this function `get_sidecar_path` https://github.com/ipetrovbg/RapidRAW/blob/98cbbafb826a82c349871447074c046da1c25679/src-tauri/src/file_management.rs#L276, so when this file exists `{}.rrdata` it adds `is_edited`.
There is just this problem with deleting this file when filters, ratings etc. get reset.

image

ipetrovbg:
This bit will also need to be extended so that it takes into account the default settings https://github.com/ipetrovbg/RapidRAW/blob/main/src-tauri/src/file_management.rs#L186

CyberTimon:
Added here: https://github.com/CyberTimon/RapidRAW/commit/d278d0cfb8ba8fb841434be0dea7251ea58f5c75

Will appear in the next release :)

image

Issue #194, opened 2025-08-01
Make zoom indicator relative to actual file size


Sterling-101:
I have a request to bring the zoom in line with other programs I have used where the zoom indicator shows the percentage you are zoomed in on the actual file size.
I find it easier to work with that absolute instead of 100% being the whole frame, 100% would be zoomed to 1:1 so if you are zooming in to check focus you can know you're at 1:! instead of over-shooting and having to guess where you are in the zoom.
Not a huge thing, but would definitely be useful.

tobiaskuntzsch:
First of all, thank you for the great application. I took a closer look at the slider. The editor shows and works with the preview. This is between 720px to 3840px according to the settings. The zoom then also applies to this image. Should the percentage display then apply to the preview size or the original resolution, which is never reached in zoom without pixelating.

Sterling-101:
My thinking (and this may go against the way the program works) is that if you go to 100% zoom then this should be a direct 1:1 of the image, not just the preview. That way you know you aren't zooming past 1:1 when confirming sharpness of details.
If you use RawTherapee this is how the zoom function works on that. Zoomed to fit will be a value below 100%, depending on your monitor size. If you then zoom to 100% you are looking at the image at the same pixel ratio as your screen, and anything above is going to be pixelated.
As it is at present I'm not sure exactly whether I'm zoomed to 1:1 and I've missed focus or similar, or I'm actually zoomed in too far and it's pixelating due to the program zooming in too far.
Hope that makes a bit more sense!

tobiaskuntzsch:
Ok, I understand. I've seen that pressing F also creates a preview in the size of the original image. One idea would be to use this function and when the zoom exceeds the resolution of the preview, automatically switch to the original image. This way you could check the focus and have the performant preview image for normal use. I'll take a look at whether this is doable.

Sterling-101:
I've just done a quick demo video of how the zoom system works in ART (a fork of RawTherapee) so you can get some idea of how I'm thinking.

https://github.com/user-attachments/assets/27cd5989-5f7f-448a-9805-d90b095c6be1

tobiaskuntzsch:
I experimented a bit yesterday, I've gotten to the point where it loads the image in original quality when zooming. However, when you change the adjustments during zooming, these aren't transferred yet. I also still need to solve a few problems with the mouse events, but in principle I believe this is doable. I hope I find some time on the weekend, then I would prepare a test version.

[...]

PR #223, opened 2025-08-07
Issue 194 relative zome indicator

tobiaskuntzsch:
Implements issue #194:
Added a relative zoom indicator that shows zoom percentage relative to the original image dimensions (10%-200%) instead of window size.

- Adjustment of zoom percentage relative to the original image size
- When the displayed image exceeds the preview size, the original is loaded
- Adjustments are also displayed while zoomed; since this is performance-intensive, a loading spinner appears
- Spacebar and double-click of the left mouse button toggle between FitToWindow, 100%, 200%
- Double-clicking the zoom slider now also performs a reset, like all other sliders
- Clicking on the percentage number opens an input field for entering the zoom level, like with the other
  sliders

 I've tested everything on Linux (except for the AI features) and haven't found any errors so far. Tests on
  Windows and Mac are still pending. Since I normally program in Ruby, I had to adjust a bit - I hope my code
  is still understandable.

### DIFF
--- a/src/components/panel/Editor.tsx
+++ b/src/components/panel/Editor.tsx
@@ -422,12 +422,17 @@ export default function Editor({
       }
     }
     
+    const doubleFitPercent = fitPercent * 2;
+    
     if (onZoomChange) {
       if (Math.abs(currentPercent - fitPercent) < 5) {
+        // Zoom 2x FitToWindows
+        onZoomChange(doubleFitPercent < 100 ? doubleFitPercent / 100 : 1.0);
+      } else if (Math.abs(currentPercent - doubleFitPercent) < 5 && doubleFitPercent < 100) {
+        // Zoom 100%
         onZoomChange(1.0);
-      } else if (Math.abs(currentPercent - 100) < 5) {
-        onZoomChange(2.0);
       } else {
+        // Zoom FitToWindows
         onZoomChange(0, true);
       }
     }
--- a/src/hooks/useKeyboardShortcuts.tsx
+++ b/src/hooks/useKeyboardShortcuts.tsx
@@ -128,7 +128,7 @@ export const useKeyboardShortcuts = ({
             ? Math.round((displaySize.width / originalSize.width) * 100)
             : 100;
           
-          // Toggle between fit-to-window, 100%, and 200%
+          // Toggle between fit-to-window, 2x fit-to-window (if < 100%), and 100%
           let fitPercent = 100;
           if (originalSize && originalSize.width > 0 && originalSize.height > 0 && baseRenderSize && baseRenderSize.width > 0 && baseRenderSize.height > 0) {
             const originalAspect = originalSize.width / originalSize.height;
@@ -142,12 +142,16 @@ export const useKeyboardShortcuts = ({
               fitPercent = Math.round((baseRenderSize.height / originalSize.height) * 100);
             }
           }
-          
+
+          const doubleFitPercent = fitPercent * 2;
           if (Math.abs(currentPercent - fitPercent) < 5) {
+            // Zoom 2x FitToWindows
+            handleZoomChange(doubleFitPercent < 100 ? doubleFitPercent / 100 : 1.0);
+          } else if (Math.abs(currentPercent - doubleFitPercent) < 5 && doubleFitPercent < 100) {
+            // Zoom 100%
             handleZoomChange(1.0);
-          } else if (Math.abs(currentPercent - 100) < 5) {
-            handleZoomChange(2.0);
           } else {
+            // Zoom FitToWindows
             handleZoomChange(0, true);
           }
           return;


CyberTimon:
Thank you so much for taking the time to contribute to RapidRAW - this looks great! I’m happy to take a closer look over the weekend.

CyberTimon:
> Adjustment of zoom percentage relative to the original image size

Thank you works perfect!

> When the displayed image exceeds the preview size, the original is loaded

This is definitely a smart approach, but it seems to have a significant impact on performance. With the original image loader, working with a 48MP image became nearly unusable. Even with smaller images (around 24MP) the UI freezes during even minor adjustments or while zooming in and out.

> Spacebar and double-click of the left mouse button toggle between FitToWindow, 100%, 200%

This & the rest is great!

---

I'm not entirely sure we should merge the full PR just yet. Loading the original image still feels too heavy at the moment, and the detection logic isn’t reliable - in many cases, it fails to load the original when it should (in my testing).

tobiaskuntzsch:
Regarding your question from the issue:

Keeping the zoom settings when switching between images would  be doable, but I can't think of a workflow where that would make sense to me. When I'm photographing, I always change the perspective, so every image is  different. First, I always want to see if I like the image crop, that's why I actually prefer it when the entire image is displayed.

Regarding performance:

Yes, it takes some time to load, especially on older, less powerful PCs I understand the problem. But without loading the full resolution, the zoom doesn't make much sense to me. Most of the time I only use zoom when I want to see if a face or object is really in focus and not blurry. If I zoom in the preview like currently, this evaluation unfortunately doesn't work because then I only see pixels. I think in most cases the benefit outweighs the performance issues, you don't actually do the main image editing in zoom mode. I created two images for comparison to illustrate what I mean.
Bildschirmfoto_20250808_224255
(The  image shows approximately 100% with the current zoom)

Bildschirmfoto_20250808_223822

(This is with the new Zoom)


One idea to improve performance would be to load the original later, but what would be a good value?

What do you mean by GUI freezing? Actually the images should be loaded asynchronously if I didn't do anything wrong. Unfortunately I don't have 48MP images, my camera only manages 24MP ;-) and the slider can still be moved quite well there, even when zoom is active. What loads afterwards is the adjustments, but I think that's also the case in Lightroom Web and can't be avoided in a web-based application.


https://github.com/user-attachments/assets/c2aac87c-d885-48ea-91fd-d91542a17deb

In this example I changed the exposure very drastically, you probably wouldn't normally do that and it feels okay to me. Does it run differently on your system?

If you perhaps have a test image where it causes problems, I can try to reproduce it and optimize it.

CyberTimon:
> Keeping the zoom settings when switching between images would be doable, but I can't think of a workflow where that would make sense to me. When I'm photographing, I always change the perspective, so every image is different. First, I always want to see if I like the image crop, that's why I actually prefer it when the entire image is displayed.

Makes sense - on Lightroom (not classic) the zoom is synced across images so I'm just a bit used to this. 

> Spacebar and double-click of the left mouse button toggle between FitToWindow, 100%, 200%

Do you still think 100% & 200% are good values to zoom to? I definitely agree with 100% but I'm not very sure about 200%, maybe it should be a 1.5x zoom which gets dynamically calculated. This can also be added later, after merging.

---

Here's the issue I have.. I have set my preview resolution to 1920px. Then I load the image I want to edit -> everything runs fine. After doing some adjustments (which are all very fast and just the preview size) I want to zoom in. I zoom in and it loads the high res image. I zoom back to FitToWindow (where it should only display the preview) and it still tries to always also update the high res image in the background (it becomes really slow as you can see, in comparison to the beginning):

https://github.com/user-attachments/assets/f5f82035-427a-4e1b-8b24-8bf83a8a2858

I hope this helps. If you want a sample image, I can give you this:
sample

I have a Ryzen 5 5600, 32GB RAM, 1TB NVMe SSD and a RTX 4070

CyberTimon:
Hey @tobiaskuntzsch 

In the meantime, we merged https://github.com/CyberTimon/RapidRAW/pull/166, which refactors the UI to TypeScript. Small adjustments and a rebase is needed for this PR. If you have any questions, feel free to ping @varjolintu or me.

Thank you!

tobiaskuntzsch:
Ok, I think I understand your problem from above, I'll look at this and the Typescript changes the days and update the pr

CyberTimon:
Amazing - looking forward to it!

tobiaskuntzsch:
I have fixed the problem so that changes are applied directly to the preview image even after a zoom, and I have migrated the code to TypeScript thanks to Claude. Unfortunately, there were so many changes in Git that a rebase was necessary so that you can recognize the actual changes. It looks good for me now, take a look and see if it works for you.

Two points are still open:
- Saving the zoom when switching between images
- Instead of 200%, only zoom to 150% with mouse and spacebar

Should I still implement this or should we leave it as is for now?

src/App.tsx
```diff
@@ -193,6 +193,54 @@ function App() {
...
+  const [previewSize, setPreviewSize] = useState({ width: 0, height: 0 });
+  const [baseRenderSize, setBaseRenderSize] = useState({ width: 0, height: 0 });
+  const [originalSize, setOriginalSize] = useState({ width: 0, height: 0 });
+  const [isFullResolution, setIsFullResolution] = useState(false);
+  const [fullResolutionUrl, setFullResolutionUrl] = useState<string | null>(null);
+  const [isLoadingFullRes, setIsLoadingFullRes] = useState(false);
+  const fullResRequestRef = useRef<any>(null);
+
+  // Track display size changes for zoom calculations
+  const handleDisplaySizeChange = useCallback((size: any) => {
```
varjolintu, line 206: You should use `ImageDimensions` interface here with `size`, imported from `useImageRenderSize.tsx`. Then you can modify the `useState`'s to `useState<ImageDImensions>({ width: 0, height: 0 });` etc.

src/App.tsx
```diff
@@ -1291,19 +1339,146 @@ function App() {
...
+      if (!selectedImage?.path) return;
+      
+      if (fullResRequestRef.current) {
+        fullResRequestRef.current.cancelled = true;
+      }
+      
+      const request = { cancelled: false };
+      fullResRequestRef.current = request;
+      
+      invoke('generate_fullscreen_preview', {
```
varjolintu, line 1354: `'generate_fullscreen_preview'` -> `Invokes.GenerateFullscreenPreview`.

src/App.tsx
```diff
@@ -1291,19 +1339,146 @@ function App() {
...
+  useEffect(() => {
+    if (isFullResolution && selectedImage?.path) {
+      setFullResolutionUrl(null);
+      setIsLoadingFullRes(true);
+      requestFullResolution(adjustments);
+    }
+  }, [adjustments, isFullResolution, selectedImage?.path, requestFullResolution]);
+
+  // Handle zoom changes with percentage relative to original image size
+  const handleZoomChange = useCallback((input: number | string) => {
```
varjolintu, line 1384: Do not use `number | string` as variable type. Use either one. A good example is the code above which doesn't make any sense.
```
let targetZoomPercent = input;
if (input === 'fit-to-window') {
```

src/App.tsx
```diff
@@ -1291,19 +1339,146 @@ function App() {
...
     setZoom(transformState.scale);
-  }, []);
+
+    if (originalSize.width > 0 && baseRenderSize.width > 0) {
+      const targetZoomPercent = (baseRenderSize.width * transformState.scale) / originalSize.width;
+      const currentDisplayWidth = baseRenderSize.width * transformState.scale;
+      const shouldUsePreview = previewSize.width > 0 && currentDisplayWidth <= previewSize.width;
+      const needsFullRes = targetZoomPercent > 0.5;
+      
+      if (needsFullRes && !isFullResolution && !shouldUsePreview) {
```
varjolintu, line 1460: This code block is duplicated.

src/components/panel/Editor.tsx
```diff
@@ -58,6 +58,14 @@ interface EditorProps {
   uncroppedAdjustedPreviewUrl: string | null;
   updateSubMask(id: string | null, subMask: Partial<SubMask>): void;
   waveform: WaveformData | null;
+  onDisplaySizeChange?(size: any): void;
+  onInitialFitScale?(scale: number): void;
+  onZoomChange?(zoom: number | string): void;
+  originalSize?: { width: number; height: number };
+  baseRenderSize?: { width: number; height: number };
```
varjolintu, line 65: `ImageDimensions` interface here too.

src/components/panel/Editor.tsx
```diff
@@ -343,6 +394,42 @@ export default function Editor({
...
+      }
+    }
+    
+    if (onZoomChange) {
+      if (Math.abs(currentPercent - fitPercent) < 5) {
+        onZoomChange(1.0);
+      } else if (Math.abs(currentPercent - 100) < 5) {
+        onZoomChange(2.0);
+      } else {
+        onZoomChange('fit-to-window');
```
varjolintu, line 428: If you really need to give `onZoomChange()` multiple values, add more parameters to the function. For example instead `'fit-to-window'` could be used with `onZoomChange(zoomValue: number, fitToWindow: boolean)` where the boolean being true, the zoom value is ignored when called with `onZoomChange(0, true)`.

CyberTimon:
Instead of a fixed zoom scale (e.g. 150%) wouldn't it be smarter to just do a 1.5x from the image size so it feels the same on every image, regardless of it's size? So fit to screen -> 1.5x -> 100%?

tobiaskuntzsch:
@varjolintu Thanks for the feedback, I've attempted to implement your recommendations."

@CyberTimon I haven't quite understood the thing with the 1.5 yet. When an image loads in Preview, it's "FitToWindow", which in my example is 25%. So a scaling factor of 0.25. If I set it to 1.0, then it's 100% and if I set it to 2.0 it's 200%. This is always relative to the image resolution. Isn't it already the way you want it?

CyberTimon:
@tobiaskuntzsch Hey, thanks for pushing the fixes and implementing the changes after the TypeScript migration! I really appreciate it.
My apologies, I see that my "1.5x" comment was a bit confusing.

My idea is only about changing the behavior of the spacebar/double-click shortcut.
Right now, the shortcut toggles between Fit to Window -> 100% -> 200%.
With a large image, "Fit to Window" might only be 25% of the original size. The jump from 25% straight to 100% can feel very abrupt and disorienting. You lose all context of where you were.

Instead of jumping to a fixed 100%, what if the first click performed a more gentle, relative zoom?

Here’s the flow I have in mind:
Start: You're at Fit to Window. Let's say for this image, that's 25%.
1st Click: Instead of going to 100%, we zoom in by a factor of 2 from the current view. So, 25% * 2 = 50%. This is a much smaller, more manageable jump to check focus.
2nd Click: Now we go to the absolute 100% for pixel-peeping.
3rd Click: We cycle back to Fit to Window.

So the new shortcut cycle would be: Fit to Window -> A gentle intermediate zoom (e.g., 2x the fit scale) -> 100% -> Fit to Window.
This makes the first zoom step feel consistent on any image, whether "Fit to Window" is 25% or 75%. The slider would still work exactly as you've designed it, showing the correct absolute percentage (e.g., 50%, 100%).
Does that make more sense?

tobiaskuntzsch:
I have integrated the ImageDimensions, merged the main branch, and resolved the conflicts.

The feature with the zoom FitToWindow -> 2x FitToWindow -> 100% has been implemented. If 2x FitToWindow would exceed 100%, this step is skipped and it zooms directly to 100% instead.

The feature for saving the zoom level when switching images is not that simple - unfortunately, I haven't found a nice, straightforward solution for it yet.

CyberTimon:
> The feature for saving the zoom level when switching images is not that simple - unfortunately, I haven't found a nice, straightforward solution for it yet.

That’s okay - we don’t need to implement this feature right now. I probably haven't said it clearly before. 

As for the rest, just let me know when you’re ready, and I’ll review it.
Also, is the bug with the high res preview being stuck in the background fixed when returning to FitToWindow?

tobiaskuntzsch:
I tested it under Linux (Kubuntu KDE) and so far I haven't been able to find any errors. The bug with the performance issue then returning to FitToWindow is fixed. If someone has Windows, feel free to test it as well. Otherwise it could be merged, i guess.

CyberTimon:
Perfect!! Thank you :) I will review this tomorrow 💯 , I don't have time today sadly.

CyberTimon:
Tested, works great! Thank you - we can merge this. If any issues arise, I can resolve them later.

CyberTimon:
@tobiaskuntzsch Sadly it introduced a quite big bug which I haven't been able to solve for the last hours. The mouse double click zoom now always zooms in at the center, instead of where we double clicked. This happens because we don't use the libraries double click zoom function anymore. Do you have a idea on how we can solve this? Thank you very much

CyberTimon:
Reverted to the old double click mouse behavoir until we find a better solution for this: https://github.com/CyberTimon/RapidRAW/commit/58aa7664d64d4c426729948db3f986934cfbbe53

tobiaskuntzsch:
oh, ok. i will take a look this weekend if i find a better solution.

CyberTimon:
Thank you!

tobiaskuntzsch:
Unfortunately, I haven't found a solution yet, but I'm still trying.

Issue #230, opened 2025-08-09
Image exports seem broken as of `2cc1cb3`


CM-IV:
OS: Arch Linux (btw)
RR Version: `2cc1cb3`
Format: Fujifilm RAF

I cannot run the release appimages, so I pull from main and build from source.

JPEG, PNG, and TIFF file exports seem to not work.  I am able to create a new directory to save them in with the file-picker, but none of the images appear there.  There is an error message that very briefly appears after going through the export file-picker.

<img width="539" height="254" alt="Image" src="https://github.com/user-attachments/assets/ccc1d4f4-753d-41ae-8899-496bc5445722" />

CyberTimon:
Probably a bug since we refactored the frontend to TypeScript. Luckily in the latest release, the export seems to work **when we don't change the JPEG quality**.

Thank you for the PR; will take a look soon.

CyberTimon:
Fixed in https://github.com/CyberTimon/RapidRAW/commit/710242e3f5488cf73a532f6b699273f6f27eb0c0. Thank you for the PR!

PR #231, opened 2025-08-09
Fix issue with String jpeg quality type being sent to Rust back-end expecting u8.

CM-IV:
This should address the issue #230 where the React form for Jpeg quality was sending a String value to the back-end.  I am  using `parseInt` to force it into an integer before being sent to Rust.

```tsx
              {fileFormat === FileFormats.Jpeg && (
                <div className="flex items-center gap-2">
                  <label className="text-sm w-20">Quality</label>
                  <input
                    className="w-full h-1 bg-surface rounded-lg appearance-none cursor-pointer accent-accent"
                    disabled={isExporting}
                    max="100"
                    min="1"
                    onChange={(e: any) => setJpegQuality(parseInt(e?.target?.value))}
                    type="range"
                    value={jpegQuality}
                  />
                  <span className="text-sm font-mono w-12 text-right">{jpegQuality}</span>
                </div>
              )}

```

### DIFF
--- a/src/components/panel/right/ExportPanel.tsx
+++ b/src/components/panel/right/ExportPanel.tsx
@@ -44,15 +44,15 @@ export default function ExportPanel({
   selectedImage,
   setExportState,
 }: ExportPanelProps) {
-  const [fileFormat, setFileFormat] = useState('jpeg');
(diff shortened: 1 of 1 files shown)


CyberTimon:
Looks good to me - thank you for the fix!

PR #234, opened 2025-08-10
Fix switching images quickly from keyboard

varjolintu:
When you open an image, and the bottom section shows images horizontally, clicking a different image causes the image to be changed immediately. But switching image from <kbd>ArrowLeft</kbd> or <kbd>ArrowRight</kbd> can be only done after the image has been fully loaded. Removing that code section allows switching images more quickly from keyboard as well.

Actually I'm not sure which way is the one you have intended:
1. Image must be fully loaded before it can be switched
2. Image can be switched even if not fully loaded

So please elaborate if this fix is not what you want :)

### DIFF
--- a/src/hooks/useKeyboardShortcuts.tsx
+++ b/src/hooks/useKeyboardShortcuts.tsx
@@ -158,10 +158,6 @@ export const useKeyboardShortcuts = ({
       }
 
       if (['arrowup', 'arrowdown', 'arrowleft', 'arrowright'].includes(key)) {
-        if (isViewLoading) {
-          event.preventDefault();
-          return;
-        }
         event.preventDefault();
 
         if (selectedImage) {


CyberTimon:
> Image must be fully loaded before it can be switched

Actually this, until I find a better solution to manage the image loading states. Currently we have the issue that when quickly switching between images, the previous one glitches above the newly loaded one (especially in a folder which contains portrait and landscape images)

varjolintu:
> > Image must be fully loaded before it can be switched
> 
> Actually this, until I find a better solution to manage the image loading states. Currently we have the issue that when quickly switching between images, the previous one glitches above the newly loaded one (especially in a folder which contains portrait and landscape images)

Maybe I could see if that is fixable too.

CyberTimon:
That would be a dream - it's quite hard I guess as I already tried to fix it so many times... but feel free to give it a shot 😃

varjolintu:
> That would be a dream - it's quite hard I guess as I already tried to fix it so many times... but feel free to give it a shot 😃

That's a really strange bug. I already tried adding a new reference to the last path selected, and even after returning when it doesn't match with the loaded image, incorrect image is displayed. Even ImageCanvas shows the correct path. So I haven't been lucky with this yet.

CyberTimon:
I think the problem is that React loses control of the image loading process when a large image causes lag. The slowdown makes it lose its reference and fall behind, so it can’t accurately control the image.

CyberTimon:
Actually let's merge this. I don't mind if it sometimes (rare) displays above the previous image for a quick time.

varjolintu:
Thanks! Hopefully a permanent solution is found for the bug in the future.

PR #281, opened 2025-08-30
More Flatpak Updates

RayJW:
I'll leave this as a draft for the time being, until the build is actually approved by the Flathub team. To sum it up:
- This now builds the Flatpak from source instead of just repackaging the `.deb` package.
- I stylized the name everywhere according to the capitalization of the project and the username.
- I included a file to specify all cargo and node packages required to build the latest, tagged RapidRAW release offline.
- Include a `.desktop` file. It would be great practice either way, Flatpak or not, to use this file for the Linux builds. That opens up the possibility of setting things like category that don't seem to be handled by the default Tauri build process.
- Edit `Cargo.toml` to actually build using the correct release instead of always outputting 0.0.0.

Let me know if any of those changes seem problematic, and we can adjust things in parallel with the Flathub submission.


RayJW:
@CyberTimon the Flathub team is now happy with the way things are. However, they want me to use the `.desktop` file and `.metainfo.xml` directly from upstream. So hopefully after this one there might be one final PR that adjusts the manifest to build from these upstream files and then that should be mostly it for an official Flatpak on Flathub.

However, those changes will have to wait until there is a new tagged release where these files are included so that I can reference them in the building process.

CyberTimon:
Will check the PR later, but quick question - why do we set this to the correct release number? With this I have to manually bump it everytime I publish a new release?

> Edit Cargo.toml to actually build using the correct release instead of always outputting 0.0.0.

RayJW:
> Will check the PR later, but quick question - why do we set this to the correct release number? With this I have to manually bump it everytime I publish a new release?
> 
> > Edit Cargo.toml to actually build using the correct release instead of always outputting 0.0.0.

So, I was wondering the same thing but in the other direction. How do the releases you create know the correct release number? Whenever I compile from source with `npx dev tauri --release` or build the Flatpak, the version number of the app will show as 0.0.0, like it's configured.

Also, from a more philosophical standpoint while it is true, that it has to be manually adjusted, realistically doing it in whatever automated way is just wrong. E.g. when people want to depend on RapidRAW in some way for their own Rust app or whatever. How would any automated version checker like Cargo, npm, or whatever, ever know what version the package is at, was at, or will be at, except for the git tags?

CyberTimon:
Right now, the version is set correctly when building with GitHub Actions:
[https://github.com/CyberTimon/RapidRAW/blob/main/.github/workflows/build.yml](https://github.com/CyberTimon/RapidRAW/blob/main/.github/workflows/build.yml)

I don’t expect RapidRAW to be used as a dependency anytime soon, so I didn’t implement version handling with that in mind. For local testing, I’m fine with the version showing as 0.0.0.

What do you think?

RayJW:
> Right now, the version is set correctly when building with GitHub Actions: [`main`/.github/workflows/build.yml](https://github.com/CyberTimon/RapidRAW/blob/main/.github/workflows/build.yml?rgh-link-date=2025-08-31T15%3A30%3A36Z)
> 
> I don’t expect RapidRAW to be used as a dependency anytime soon, so I didn’t implement version handling with that in mind. For local testing, I’m fine with the version showing as 0.0.0.
> 
> What do you think?

I was just typing out an edit when your response came in :D I just saw, that it is handled in the pipeline. I'd personally adjust the version number the intended way. That way, you can be sure that it will always be correct, even when the Tauri or package specific build processes might change.

However, ultimately it's up to you but from experience I'd take adjusting a number in 3 different places every few weeks and documenting the process, over maintaining additional complexity for every pipeline etc. any day, but that's just me. I'm totally okay with reversing it if you prefer it the current way, at the end of the day it is you doing most of the work for RapidRAW!

If you decide to adjust the version number, I would recommend adding a small release markdown in a docs folder and documenting a small checklist you follow for every release. That way it's easier, any way to make sure you never miss anything.

RayJW:
Hey @CyberTimon, I took another look at [the docs](https://tauri.app/reference/config/#version), and it seems like the version number in `Cargo.toml` and `package.json` are just fallbacks. The supported way of managing your version number, is setting it in `tauri.conf.json`. If you want, I think we could delete the version field altogether, since you don't expect/support RapidRAW as a library for another program. Then there would be no weird version numbers present and there is only one field you have to adjust for new versions, and all the complexity related to setting the version number in all the pipelines could be deleted. Alternatively, we could keep the other versions at `0.0.0`, while only keeping the `tauri.conf.json` at the correct version.

How would you like to move forward?

CyberTimon:
If this works, we can remove it and just keep tauri.conf.json up to date. That way, I only need to manage one version number file. Will review this later, when I get back home :)

RayJW:
> If this works, we can remove it and just keep tauri.conf.json up to date. That way, I only need to manage one version number file. Will review this later, when I get back home :)

> If you plan to publish your package, the most important things in your package.json are the name and version fields as they will be required. The name and version together form an identifier that is assumed to be completely unique. Changes to the package should come along with changes to the version. If you don't plan to publish your package, the name and version fields are optional.

[Source](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#version)

> This field is optional and defaults to 0.0.0. The field is required for publishing packages.

[Source](https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field)

In case you don't plan to publish neither the Node part nor the Rust part, it should be unproblematic to get rid of both fields :)

CyberTimon:
I’ll merge this soon. Just to confirm - this won’t conflict with the current GitHub build workflow, right?https://github.com/CyberTimon/RapidRAW/blob/main/.github/workflows/build.yml

From what I understand, the workflow would overwrite it again. Is that correct? (we don't have to remove this if it doesn't conflict / fail)

Thanks a lot!

RayJW:
> Just to confirm - this won’t conflict with the current GitHub build workflow, right?

I don't see, why it should. The build process doesn't change at all for those other packages. Although, it might make sense to remove that unnecessary complexity down the line, they will keep overwriting the version number with the tag, like they're doing now.

CyberTimon:
Thank you - looks good to me 👍

PR #284, opened 2025-08-31
Transition to Rust 2024 Edition

RayJW:
This PR updates the project to use the newer 2024 edition for Rust. This, for one, enables the usage of newer Rust features, but in the meantime more importantly just ensures that formatting etc. adheres to the most up-to-date best practices in writing Rust code.

Most of the changes are fully automated and thoroughly tested by the Rust project. You can see the migration guide I followed [in the docs](https://doc.rust-lang.org/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html). The only manual changes were made to `comfyui_connector.rs` to fix three warnings for [new temporary scope rules](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/temporary-tail-expr-scope.html).

In my testing this didn't affect anything, but of course if more people could look at these changes, that would be great.

PR #286, opened 2025-08-31
Bump All Major Dependencies

RayJW:
With this PR I tried to modernize the whole codebase a bit by upgrading all dependencies including major jumps like for wgpu.

It is no secret, that for now the performance of RapidRAW can be a bit lacklustre compared to other alternatives. This definitely seems to help a bit on the two systems I've tested these changes on. One with an AMD card and one with an Intel iGPU. There are more things I'd like to explore, but I think this change together with #284 should lay a good foundation to explore more performance optimizations. I particularly want to look at wgpu caching for pipelines, which is now available thanks to these new versions.

Before merging this, it certainly needs some more testing across more devices, and it would probably make sense to merge #284 first and then reformat everything according to the new standard. But I still wanted to create this PR for now, so testing can be done and a review of the logical changes can be done.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "dependencies": {
         "@dnd-kit/core": "^6.3.1",
         "@dnd-kit/utilities": "^3.2.2",
-        "@tauri-apps/api": "^2.5.0",
+        "@tauri-apps/api": "^2.8.0",
         "@tauri-apps/plugin-dialog": "^2.2.2",
         "@tauri-apps/plugin-opener": "^2",
         "@tauri-apps/plugin-os": "^2.3.0",
@@ -38,7 +38,7 @@
       },
       "devDependencies": {
         "@eslint/js": "^9.32.0",
-        "@tauri-apps/cli": "^2.5.0",
+        "@tauri-apps/cli": "^2.8.3",
         "@types/lodash.debounce": "^4.0.9",
         "@types/react": "^19.1.8",
         "@types/react-dom": "^19.1.6",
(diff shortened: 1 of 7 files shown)


CyberTimon:
Thank you for the PR 🥇! I’ll give it a full review when I get home, as mentioned in #284, but from a quick glance it looks good.

CyberTimon:
From a quick test during the last break, I noticed that the thumbnails seem to lag by one adjustment. It could just be a one-off bug on my MacBook, but I’ll dig into it more thoroughly when I get home. Do you have any idea what might be causing this?
EDIT: Also when zooming in, which loads the full screen high res preview, the application crashes.

RayJW:
> From a quick test during the last break, I noticed that the thumbnails seem to lag by one adjustment. It could just be a one-off bug on my MacBook, but I’ll dig into it more thoroughly when I get home. Do you have any idea what might be causing this? 

Yep, I missed that. I narrowed it down to the version bump from wgpu 24.0 to 26.0 (25.0 throws a bunch of errors, I'm not sure where they're coming from). Looking at the changelog from wgpu I'm not sure what could be causing this, but I'm still looking into it.

> EDIT: Also when zooming in, which loads the full screen high res preview, the application crashes.

This doesn't seem to happen on my device. Not sure if it's down to macOS? If you have any more pointers, let me know.

CyberTimon:
Thanks! I’ll also search for a fix when I have some time. My driving lessons were a bit long today, so I couldn’t look into it yet, but I’ll get to it tomorrow.

CyberTimon:
Hey @RayJW, I've found the fix for both the crash and the thumbnail lag! It was a race condition caused by the wgpu update; changing `device.poll(wgpu::PollType::Poll)` to the blocking `device.poll(wgpu::PollType::Wait)` in `read_texture_data` resolves everything.

CyberTimon:
I’ll review a few other functions later, but for now we’re all set.

CyberTimon:
Tested the other parts - they look good to me. We can fix upcoming issues later :)

Thank you again for the PR 💯

RayJW:
> Hey @RayJW, I've found the fix for both the crash and the thumbnail lag! It was a race condition caused by the wgpu update; changing `device.poll(wgpu::PollType::Poll)` to the blocking `device.poll(wgpu::PollType::Wait)` in `read_texture_data` resolves everything.

Wow, I looked at that specifically yesterday and totally overlooked it. Should've slept over it first. Thanks for finding a fix! Tested everything again and seems fine now :) Thanks for moving this forward!

PR #308, opened 2025-09-10
Cargo style documentation for developper

slcorrensi:
This is a beginning of code comments to create cargo style documentation for the backend. It would allow easier dive into the project for newcomers. Not all the files are yet commented (especially ai_processing.rs, comfyui_connector.rs and file_management.rs). Some function explanations are not the best yet but this is a start.


CyberTimon:
Thanks for putting this together! I’d prefer not to merge it right now though, since I find comments slow me down in development a lot (as I use a lot of AI to code). Maybe we can revisit a different approach to docs later on. What do you think?

da-anda:
IMHO the best place for a documentation is within the code and AI tools should not have a problem with adding documentation along the way. They can also use the documentation to understand what a method is supposed to do.

sidenote: you should probably add an `AGENTS.md` file as well as an `.editorconfig` file to ensure certain CGL, regardless if code was created by AI or a dev

kellpossible:
> IMHO the best place for a documentation is within the code and AI tools should not have a problem with adding documentation along the way. They can also use the documentation to understand what a method is supposed to do.

Seconding this!. Also sometimes its nice to know the intended purpose of a function or data type, in isolation from the rest of the system, I think this is where doc comments come in. Otherwise it's more likely to get used and changed in strange ways. 

Also aside from whether or not AI will take any notice of the comments (I think it will too though), I think some docs like this can help and encourage human contributors 🙂

Issue #307, opened 2025-09-10
When opening a Folder with 2000+ .dng files the application is hogging full CPU


mRs-:
Hey there,

I'm a concert Photographer and for me the Application is not usable, when opening a folder with one concert and 2000+ .dng files the Application is hogging all my CPU and it's not usable at all.

Maybe it would be a good opportunity to optimize for a big folders so that only a small portion of images is being processed instead of all the images at the same time?

anonimodwalsu-prog:
Same

riccio8:
@mRs- @anonimodwalsu-prog  Does this only happen with uncompressed or slightly compressed images (tiff, dng, raw) or also with many images in lighter formats (such as jpeg, png, etc.)?

anonimodwalsu-prog:
Hey, for me it happens whenever I load files with more than say around 30 pics, after some time it works, it seems to load everything in a file at once, which does not seem very resource efficient, maybe a good idea would be to only load some lower res previews of pics and after clicking on one load that one in full res as well as 2 to the right and 2 to the left of the clicked on pic, such would allow to drastically reduce load on the cpu and make everything run a lot smoother.

anonimodwalsu-prog:
Oh and it happens when using uncompressed Fuji raw, jpegs from my xt3 and a6500 as well as the raws from the a6500. I use an older 7th gen quad core i7 with 8threads and 24gb of ram

riccio8:
Are you telling me that even with only 30 images, your computer freezes on this screen? Or does the program crash anyway? Before loading the image (which is done by double-clicking), the program uses thumbnails, which take up much less space and are always mapped in memory.


<img width="3839" height="2076" alt="Image" src="https://github.com/user-attachments/assets/80b148b0-0b30-44de-a318-12434b18de40" />

anonimodwalsu-prog:
It does not freeze per say, i would say it becomes extremely laggy as it is still usable but loads a selected image after a couple seconds. While checking task manager I saw that the cpu was running at 100%, the moment I open the file with the pics, for a couple minutes. It sometimes freezes for a little bit but always becomes responsive after waiting around 10 seconds, never crashed.

riccio8:
It's probably a problem caused by the function defined here:
https://github.com/CyberTimon/RapidRAW/blob/main/src-tauri/src/file_management.rs#L584

This function uses all available threads and GPUs if present, then passes each thumbnail created to Tauri.

One solution could be to do something like this:



```rust
[...]

Issue #317, opened 2025-09-13
RAW Image preview is wrong


svatas:
I am using 1.3.12 in Linux (Tuxedo OS on KDE Plasma 6 and Wayland). I have a folder with RAW images from my Sony A7 III (arw).

Step 1: When I see the folder I can see rigth amount of images in there, but preview images are for some of them completely wrong.

<img width="2696" height="1515" alt="Image" src="https://github.com/user-attachments/assets/8b362f36-4fbb-4a6b-b59d-bcf9317c3a11" />

The images 1, 5 and 6 shows the same goat I shooted when visited ZOO with kids. But in this folder you can see some illustrative images to some technical article, so there is a charger and some light taken with various camera settings. No goat shooted.

Step 2: I just switched to the folder and you can see, that preview image and the image was refreshed to the proper one, so image 1 is refreshed, but images 5 and 6 are still wrong.

<img width="2696" height="1515" alt="Image" src="https://github.com/user-attachments/assets/6ac8e8dc-5ec0-4bd1-bfeb-dc7c2a24a010" />

Step 3: same efect for image 5, preview and image refreshed and replaced by the proper one

<img width="2696" height="1515" alt="Image" src="https://github.com/user-attachments/assets/00031235-dd5e-4489-8bed-50e08ea21c62" />

Step 4: When I was quick, I can see even the wrong image and preview as I click on the image 6. For some time, until refreshed, image is there.

<img width="2696" height="1515" alt="Image" src="https://github.com/user-attachments/assets/22b62793-b1e3-4b29-9a8a-45fb3865628b" />

Step 5: Image refreshed and replaced with proper one.

<img width="2696" height="1515" alt="Image" src="https://github.com/user-attachments/assets/5461c3bd-cec6-44fb-b7d6-de5c0693b8d3" />

Step 6: In folder view everything is updatedto proper images.

<img width="2696" height="1515" alt="Image" src="https://github.com/user-attachments/assets/b0f9770f-ec7a-4319-a4cc-f3a19a829519" />

For me this looks like some refresh/update process "failure" as it is initiated when the image is previewed. So functionality is there but the trigger for update is somehow wrong.

CyberTimon:
Okay this is very strange. Can't reproduce and never happend on any system I'm aware off. But thank you for reporting.

Might be related: https://github.com/CyberTimon/RapidRAW/issues/306

Have you tried to launch RapidRAW with following commands?

```bash
[...]

PR #319, opened 2025-09-14
Optimization for generating thumbnails

riccio8:
Related to #307 generate_thumbnails

- Added the use of mimalloc (Microsoft allocator) as the global allocator
- Use of rayon ThreadPoolBuilder instead of std Thread

This is a first commit, I plan to bring many more optimizations


CyberTimon:
Thank you, this looks great! I’m planning to release the next version later today. Do you want me to go ahead and merge this now, or are you planning to add more optimizations to the PR?

riccio8:
> Thank you, this looks great! I’m planning to release the next version later today. Do you want me to go ahead and merge this now, or are you planning to add more optimizations to the PR?

If u want u can merge this in the next release, while I add more optimizations.

Thx a lot

CyberTimon:
I just tested your PR, and from what I can tell, it generates all the thumbnails in the background and only displays them once they’re all ready. In my experience, this approach introduces more lag compared to progressively generating and showing thumbnails as they become available. Am I missing something?

riccio8:
> I just tested your PR, and from what I can tell, it generates all the thumbnails in the background and only displays them once they’re all ready. In my experience, this approach introduces more lag compared to progressively generating and showing thumbnails as they become available. Am I missing something?

In reality, the logic remains the same; the only thing I changed is the type of thread pool. Instead of using the one from the std lib, I use the one from rayon, and I configure it to be used without saturating all the cores (which slows down tauri rendering). I also decided to use `mimalloc`, the memory allocator developed by Microsoft as a global allocator, which is much more efficient than the system ones and, above all, cross-platform.

So basically, the logic remains the same, but I don't saturate all the CPUs, which would block rendering until they finish, and I use an allocator optimized for storing large amounts of heavy data in memory.

It's not a game changer; in fact, I plan to make other changes, but it's a start.
Did it seem better or worse than the version you are currently using?

Thx a lot for ur feedback

CyberTimon:
Hey!

I noticed that while the core logic wasn’t changed, with your PR the thumbnails only appear once all of them finish loading, which blocks the UI until everything is ready.

On the main branch (without your PR), the thumbnails load progressively - they appear smoothly as they’re generated - and the UI stays responsive.

I can take another look in the next few days, but for now it feels like the current PR behavior is actually a step back.

Thanks!

riccio8:
> Hey!
> 
> I noticed that while the core logic wasn’t changed, with your PR the thumbnails only appear once all of them finish loading, which blocks the UI until everything is ready.
> 
> On the main branch (without your PR), the thumbnails load progressively - they appear smoothly as they’re generated - and the UI stays responsive.
> 
> I can take another look in the next few days, but for now it feels like the current PR behavior is actually a step back.
> 
> Thanks!

That's right, I took a closer look at rayon's install function and it's a blocking function, so while thread progressively passed the thumbnails created to the UI, poll.install waits for the generation to finish, which is slower.

To combine the benefits, I thought we could merge the two logics, using poll.install inside a separate thread. It's a small change, but it should make loading smoother.

Obviously, I'll work on other possible optimizations, and I hope to be able to merge them into the release after the next one.

Let me know what you think. Would that be okay?

riccio8:
Regarding the installation documentation, you can find more information here: 
[rayon install](https://docs.rs/rayon/latest/rayon/struct.ThreadPool.html#warning-thread-local-data)

CyberTimon:
I now have the problem that it emits the same file many times and displays the thumbnails duplicated:

This is your PR:
image

And this is main:
image

Do you have a idea why this might happen?

CyberTimon:
I try to fix it - give me some minutes.

CyberTimon:
Related: https://github.com/CyberTimon/RapidRAW/issues/317

CyberTimon:
I couldn't find a fix yet. Will take a look again in the next days - feel free to try resolve the issue.

riccio8:
> I couldn't find a fix yet. Will take a look again in the next days - feel free to try resolve the issue. 

Sure, I'll work on it in the next few days. It's still unstable, so I recommend not merging with the next release. Maybe I have a better option, a static thread, and then call its .spawn function, which isn't blocking, unlike .install. 

I noticed that the proposed change (os threads and rayon pool) would also have the same problem, so I'll try to fix that.

About the file duplication issue it's really weird, I saw that you tried to add a hash for the photos so you know which thumbnails were generated, very strange as a behavior

riccio8:
Which OS gives this problem?

CyberTimon:
I tried it on Windows 11 :)

riccio8:
Hey @CyberTimon 
I made a few changes:

- I replaced the install function with spawn, which is non-blocking
- After reading some issues where sudden crashes occur (such as [issues #322](https://github.com/CyberTimon/RapidRAW/issues/322)), I implemented an auto-log system which generates a log on stderr and in a file called RapidRaw.log every time there is a panic or undefined behavior

Let me know what you think, thanks :)

riccio8:
@CyberTimon done, I tested it

aNNiMON:
@riccio8 

> * I implemented an auto-log system

I've tried freshly built artifact and upon crashing there is nothing in the logs. `RapidRaw.log` literally 0 bytes

riccio8:
> @riccio8 
> 
> > * I implemented an auto-log system
> 
> I've tried freshly built artifact and upon crashing there is nothing in the logs. `RapidRaw.log` literally 0 bytes


I set only error as filter, I'll change it into Debug or Info for capturing more details

aNNiMON:
@riccio8 what about setting logging level based on the env variable `RUST_LOG` with the error level as a default? So that I can set `RUST_LOG=debug` and see detailed results.

CyberTimon:
@riccio8 Just tested the PR and everything is like 500x times slower? Is this because of the debugging? Because I see a lot of debugging in the console, but the application itself (thumbnail loading, image editing etc., is like 500 times slower)

https://github.com/user-attachments/assets/1f233a1f-fa86-47c0-9c3c-eb4a29e97716

riccio8:
> @riccio8 Just tested the PR and everything is like 500x times slower? Is this because of the debugging? Because I see a lot of debugging in the console, but the application itself (thumbnail loading, image editing etc., is like 500 times slower)
> 
> https://github.com/user-attachments/assets/1f233a1f-fa86-47c0-9c3c-eb4a29e97716
> 
> 

Yeah, that's because of the Debug filter settled as level of log. I'll change following @aNNiMON 's suggestions, Error should be a lot more fluid as experience.

This afternoon I'll edit and text.

By the way, do you think a directory with tests would be useful?

@CyberTimon @aNNiMON

riccio8:
Done, is it better?
@CyberTimon @aNNiMON

CyberTimon:
Will test it out in the coming days. Sorry I haven't had much time lately. Thank you :)

riccio8:
> Will test it out in the coming days. Sorry I haven't had much time lately. Thank you :)

Np, let me know

riccio8:
Hey @CyberTimon @aNNiMON 
I was thinking that mapping files in memory wouldn't be a bad idea, as you map 1000 at a time, and then pass the mapping reference to progressively generate thumbnails. You can do this using the crate https://docs.rs/memmap2/latest/memmap2/.

What do you think?

CyberTimon:
That sounds interesting! My knowledge of mappings is somewhat limited, but feel free to give it a try 💯. In the meantime, I’ll go ahead and merge this, since the logging and related improvements are definitely things I’d like included in the next release.

riccio8:
> That sounds interesting! My knowledge of mappings is somewhat limited, but feel free to give it a try 💯. In the meantime, I’ll go ahead and merge this, since the logging and related improvements are definitely things I’d like included in the next release.

I'll let you know. I don't have much time these days, but I'll work on it. Would you prefer me to open a _pull request_ or an _issue_ and work on it?

riccio8:
@CyberTimon 

Sorry for mentioning, but I implemented it, tested it, and it seems to perform better. I tried it on a 4TB NAS in RAID and loaded a thousand images. It took a little while at first, 2-3-4 seconds, and then loaded them all.

How do u want me to pull the request? thx

CyberTimon:
Yes that would be great! Thank you for all the work :)

riccio8:
here it is: https://github.com/CyberTimon/RapidRAW/pull/347#issue-3448803306
@CyberTimon

PR #326, opened 2025-09-16
Update App Identifier to New Domain

RayJW:
This mainly adjusts the Flatpak packaging data to the new reverse-domain structure with the WIP website. However, some invalid domains across the repo were fixed as well. I also adjusted the data for the new v1.3.13 release.

I also took the liberty of prematurely adding the to-be domain for the Flatpak package, and my editor took care of some formatting issues. Obviously, some projects don't like unrelated changes like that, so I'll revert them if wanted.


RayJW:
Once this is merged and in a tagged release, https://github.com/flathub/flathub/pull/6941 will build properly and can move ahead with publishing.

CyberTimon:
Currently, RapidRAW stores its downloaded models, presets, thumbnails, and configuration files in:
 - C:\Users\USER\AppData\Roaming\com.rapidraw.app
 - C:\Users\USER\AppData\Local\com.rapidraw.app

From what I understand, changing the identifier in the Tauri config would also change these storage paths. This would force users to regenerate thumbnails, re-download AI models, and lose all existing presets. Is that change really necessary?

Additionally, I’ve noticed a potential DNS issue: https://www.getrapidraw.com/ is only accessible with the www subdomain. Accessing it directly via https://rapidraw.com doesn’t work. Not sure if this could cause problems in combination with Flatpak, but I wanted to inform you.

RayJW:
> From what I understand, changing the identifier in the Tauri config would also change these storage paths. This would force users to regenerate thumbnails, re-download AI models, and lose all existing presets. Is that change really necessary?

I see, but from what I understand you do not have control over rapidraw[dot]com, and it would generally be pretty bad practice from a security standpoint to potentially mislead your users with domains that could be taken over by a malicious actor, as it is currently for sale. Since the current configuration is a spec-compliant [reverse-domain notation](https://en.wikipedia.org/wiki/Reverse_domain_name_notation) that would strongly imply, to users and automatic tools, that rapidraw[dot]com is the official domain of this project. I get that this might be bad from a user experience standpoint, but I feel like postponing this change is trouble waiting to happen.

> Additionally, I’ve noticed a potential DNS issue: [getrapidraw.com](https://www.getrapidraw.com/) is only accessible with the www subdomain. Accessing it directly via [rapidraw.com](https://rapidraw.com) doesn’t work. Not sure if this could cause problems in combination with Flatpak, but I wanted to inform you.

I saw that as well, but since getrapidraw.com redirects to www.getrapidraw.com, I don't see this being a problem. Is there any reason you host your website on the www subdomain? Other than it being an age-old tradition on the internet, there is no benefit to using prefixing www unless you want to use the root of the domain for something else.

CyberTimon:
Thanks for explaining - I wasn’t aware of the risks with the domain. You’re right, it’s better to make the change now than risk issues down the line. I’m fine with moving forward on this.

> Is there any reason you host your website on the www subdomain?

Yes - at the moment I can’t point the root A record to my webserver, so I can only use a CNAME (www). That’s why the site is currently only accessible via the www subdomain.

Merci

CyberTimon:
@RayJW It looks like we’ve run into an issue: https://github.com/flathub/flathub/pull/6941#issuecomment-3300244833

Flatpak requires me to host a verification .txt file on the domain root, but I can’t point the A record directly since my webserver doesn’t yet have a static IP. At the moment, I’m only able to serve the file under the www subdomain: https://www.getrapidraw.com/.well-known/org.flathub.VerifiedApps.txt

william-sy:
> @RayJW It looks like we’ve run into an issue: [flathub/flathub#6941 (comment)](https://github.com/flathub/flathub/pull/6941#issuecomment-3300244833)
> 
> Flatpak requires me to host a verification .txt file on the domain root, but I can’t point the A record directly since my webserver doesn’t yet have a static IP. At the moment, I’m only able to serve the file under the www subdomain: https://www.getrapidraw.com/.well-known/org.flathub.VerifiedApps.txt

Not sure what hosting you do, But usually a verification is done at dns level with a TXT record.

As for the website. Might I suggest you build it statically ? then you can just host it on cloudflare pages for free its not like you use a contact form or anything. 

From there hosting txt files should be easier aswell, and you can host the website on github in a private repo :)

CyberTimon:
@william-sy Thanks for the recommendation! I’ll definitely check it out soon. In the meantime, I plan to migrate the website from the current server (which doesn’t have a static IP) to a Hetzner VPS. That way, I’ll have a static IP and can properly point the A record.

SimonIT:
You can also use GitHub Pages for a static web page with your own domain

PR #334, opened 2025-09-18
Flatpak Updates & Documentation

RayJW:
This supersedes #326 by adopting all the discussed changes except the new app identifier, since the Flathub team would like to keep the current one, so it'll be the easiest to maintain that one across all platforms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,13 +53,13 @@ RapidRAW is still in active development and isn't yet as polished as mature tool
 
 *   **2025-09-17:** Automatic image culling (duplicate & blur detection)
 *   **2025-09-14:** Grid previews in community panel & improved ComfyUi workflow
-*   **2025-09-12:** New community presets panel to share & showcase presets 
-*   **2025-09-10:** Extended generative AI roadmap & started building RapidRAW website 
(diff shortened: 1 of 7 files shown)


CyberTimon:
So, we’re not merging #326? Should I go ahead and close it, then? If so, I’ll proceed with merging this one. Thanks!

RayJW:
> So, we’re not merging #326? Should I go ahead and close it, then?

I think so since, as the Flathub team explained, it's mainly a cosmetic thing it's really fine to just stick with this one as long as nothing big occurs, like a hostile takeover of this GitHub repo or whatever.

CyberTimon:
I will merge this soon and then instruct the users to either rename the folder in the appdata directory, or redownload the models / regenerate the thumbnails.

RayJW:
> I will merge this soon and then instruct the users to either rename the folder in the appdata directory, or redownload the models / regenerate the thumbnails.

Sounds good, before you tag a new release, could you let me know or if you don't mind quickly update the metainfo file with the release notes as well? Otherwise, the tagged release will use the changelog of the previous version and any instructions to rename won't be visible.

CyberTimon:
Yes of course - I can do that. Thanks for the info.

PR #335, opened 2025-09-19
Minor community guideline updates

william-sy:
Suggestion for contribution guidelines.

Feel free to decline / remark / improve.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# Owner for everything in the repo
+* @CyberTimon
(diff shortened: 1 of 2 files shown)


CyberTimon:
LGTM - but note that

> Once support is added in rawler they will flow upstream into RapidRAW

is a bit vague. The merge-back process from upstream into RapidRAW is not yet fully implemented, as discussed here: https://github.com/CyberTimon/RapidRAW/issues/336.

william-sy:
Ah oke, then we could frame it something  like: 
`Once support is added in rawler please create a ticket for RapidRAW so we can update the packages`

CyberTimon:
I switched the text from 'we' to 'I', replaced 'tickets' with 'issues' (since that's the more common term on GitHub), and made a few small wording adjustments.

PR #353, opened 2025-09-28
Fix Metainfo Formatting

RayJW:
Small fix, just saw by chance that the changelog you added for the upcoming release would stop the Flatpak from being built. The formatting for the release tag in metainfo file prohibits any tags that are not either `p`, `ol`, `ul`, `li`, `em`, or `code`. Just quickly removed them, unless you would like them replaced with `em` for emphasis?


CyberTimon:
Oh I see - thank you! I wasn't aware of this.

RayJW:
> Oh I see - thank you! I wasn't aware of this.

No problem! I should've warned you about that limitation '^^ The same should also be done across the new text you added right as I was creating the PR. Didn't see there was new stuff :)

CyberTimon:
Done, thanks!

Issue #362, opened 2025-10-01
Crash With Newest Release


RayJW:
### What happened?

With the newest release, the Flatpak is now reliably crashing. It seems like this is most likely an issue with the newly introduced logging system.

### What did you expect to happen?

No crashing.

### How can we reproduce this?

1. Install the Flatpak.
2. Start RapidRAW and open a folder.

### Operating System

Other Linux

### RapidRAW Version

v1.4.1

### Additional information

[app.log](https://github.com/user-attachments/files/22639187/app.log)

### Similar issues

- [x] I have searched existing issues and didn't find a duplicate

CyberTimon:
@riccio8 Please take a look here, seems related to https://github.com/CyberTimon/RapidRAW/pull/347

prokoudine:
Confirmed

riccio8:
> ### What happened?
> 
> With the newest release, the Flatpak is now reliably crashing. It seems like this is most likely an issue with the newly introduced logging system.
> 
> ### What did you expect to happen?
> 
> No crashing.
> 
> ### How can we reproduce this?
> 
> 1. Install the Flatpak.
> 2. Start RapidRAW and open a folder.
> 
> ### Operating System
> 
> Other Linux
> 
> ### RapidRAW Version
> 
> v1.4.1
> 
> ### Additional information
> 
> [app.log](https://github.com/user-attachments/files/22639187/app.log)
> 
> ### Similar issues
> 
> - [x] I have searched existing issues and didn't find a duplicate

Can you reproduce the error setting Debug or Info as log level?

riccio8:
> ### What happened?
> 
> With the newest release, the Flatpak is now reliably crashing. It seems like this is most likely an issue with the newly introduced logging system.
> 
> ### What did you expect to happen?
> 
> No crashing.
> 
> ### How can we reproduce this?
> 
> 1. Install the Flatpak.
> 2. Start RapidRAW and open a folder.
> 
> ### Operating System
> 
> Other Linux
> 
> ### RapidRAW Version
> 
> v1.4.1
> 
> ### Additional information
> 
> [app.log](https://github.com/user-attachments/files/22639187/app.log)
> 
> ### Similar issues
> 
> - [x] I have searched existing issues and didn't find a duplicate

Can you try this please?

https://github.com/CyberTimon/RapidRAW/actions/runs/18197442061#artifacts

RayJW:
[...]

PR #366, opened 2025-10-02
Patch for crash in v1.4.1 using flatpak 

riccio8:
This is the first patch for the error described in issue https://github.com/CyberTimon/RapidRAW/issues/362 

It introduces a more idiomatic way of handling errors and greater controls over the memory mapped file


CyberTimon:
Thank you - LGTM. Let's hope it fixes the Flatpak crash. @RayJW

PR #388, opened 2025-10-13
fix: prevent overscrolling on html element

ellie:
This PR fixes the bouncing overscroll behaviour present on macOS

Previous behaviour:

https://github.com/user-attachments/assets/64bc892a-3667-4e21-9d3e-f4fa19e01b9f

I also changed the <title> tag, not that it's displayed anywhere I could find.


CyberTimon:
Amazing - thank you for the fix!

PR #392, opened 2025-10-16
Vite 7 Update

RayJW:
I wanted to modernize the Node part of RapidRAW to explore some possible performance updates with regard to Linux. I thought I'd start with Vite and then proceed to React 19 which I'll mix with other dependency bumps etc. This seems to be working fine so I wanted to separate this for now as this shouldn't really affect the app itself.


CyberTimon:
LGTM & thank you for bumping the dependencies. Tested on Windows / macOS.

PR #393, opened 2025-10-16
React 19 Update

RayJW:
As promised, I'll try to see if I can update to React 19 including some dependencies. As React is known to create an unnecessary amount of DOMs, which especially causes problems with WebkitGTK, I hope this could help out performance on Linux a bit, but I think more work will be necessary. I'll also try to get rid of some old and obscure dependencies that aren't being used and modernizing everything else. As always, testing on all platforms is welcome with big changes like this.


RayJW:
I think this is ready now, seems like no changes were **required** for React 19, although some should probably be made in order to take advantage of the available improvements that aren't under-the-hood. Next up: Tailwind as v4.0 should probably make things a lot simpler as well :)

CyberTimon:
Looks good to me - works fine on both Windows & macOS. Thank you!

PR #394, opened 2025-10-16
TailwindCSS v4 Update

RayJW:
This is my initial attempt at migrating to TailwindCSS v4. While their upgrade tool took care of most of the work, there are definitely big differences here. However, thanks to that transition, we can now use the Vite plugin and can get rid of PostCSS. This simplifies the structure quite a bit and is supposedly faster according to the Tailwind team. I'll try to get the looks back to what they were when I find time.

But maybe you can butt in here @CyberTimon since you probably have a much better understanding of where things are and how you would want them to look. No obligations obviously, not trying to force any work on you that you don't want :)

[Migration guide for reference](https://tailwindcss.com/docs/upgrade-guide)

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -2553,7 +2553,7 @@ fn apply_window_effect(theme: String, window: impl raw_window_handle::HasWindowH
                 .expect("Failed to apply acrylic effect on Windows 11");
         } else {
             window_vibrancy::apply_blur(&window, color)
-                .expect("Failed to apply blur-sm effect on Windows 10 or older");
+                .expect("Failed to apply blur effect on Windows 10 or older");
         }
     }
 
--- a/src-tauri/src/tagging_utils/candidates.rs
+++ b/src-tauri/src/tagging_utils/candidates.rs
@@ -24,7 +24,7 @@ pub const TAG_CANDIDATES: [&str; 590] = [
     "crying",
     "thinking",
     "silhouette",
-    "shadow-sm",
+    "shadow",
     "hands",
     "feet",
     // --- Animals ---
@@ -507,7 +507,7 @@ pub const TAG_CANDIDATES: [&str; 590] = [
     "macro",
     "close-up",
     "long exposure",
-    "motion blur-sm",
+    "motion blur",
     "light painting",
     "bokeh",
     "depth of field",


CyberTimon:
Yeah - right now, everything seems broken - especially the font colors and the button/surface background colors. I’ve also noticed that image editing is significantly laggier than before, though I’m not sure why.

CyberTimon:
> I’ve also noticed that image editing is significantly laggier than before, though I’m not sure why.

After some debugging it seems like it started to become a lot slower since upgrading to React 19 and has nothing to do with this PR.

RayJW:
> > I’ve also noticed that image editing is significantly laggier than before, though I’m not sure why.
> 
> After some debugging it seems like it started to become a lot slower since upgrading to React 19 and has nothing to do with this PR.

Huh, that's weird. I might've not noticed because React is already pretty bad with performance on Linux, so I don't think it caused a major difference on my end. But it seems odd that just changing the React version would cause that. Would you mind maybe bisecting if it was any specific commit? Perhaps one of the other dependencies broke something on the other platforms?

CyberTimon:
Yes. I’m currently debugging it. I’ve also just started noticing it by editing an image, including dragging curves and making adjustments.

CyberTimon:
@RayJW This commit causes the slowdown / lag: https://github.com/CyberTimon/RapidRAW/commit/72890006808f1909cbad81cb9efa094377723c54

Downgrading vite back to version 6 completely resolves the performance issue on windows.

EDIT: After further testing, it seems the issue might be related to my environment, as the lag appears intermittently - it’s pretty random. I’m continuing to investigate this further.

CyberTimon:
Yes it's probably my system or just in the development area. The final github actions build work fine. Most likely it has nothing to do with your dependency bumps - sorry for the confusion.

RayJW:
> Yes it's probably my system or just in the development area. The final github actions build work fine. Most likely it has nothing to do with your dependency bumps - sorry for the confusion.

Huh, that's very weird. Might be some caching issue or similar.

On another note, I've found [this issue](https://github.com/tailwindlabs/tailwindcss/discussions/16517) which might help as the previous configuration was pretty heavy on `tailwind.config.js` which was migrated and seems like it might be the culprit as it is especially the dark mode also broken atm.

RayJW:
I'll close this, as too much has changed in the meantime for this to still make sense without starting over.

PR #414, opened 2025-10-26
Crash prevention and automatic restart

riccio8:
Implement crash recovery mechanism: Seeing the numerous issues where the app often crashes, while trying to resolve the problem, the app now monitors its own crashes and automatically attempts to restart after an unexpected termination. To prevent endless crash loops, a maximum threshold of 3 consecutive crashes has been implemented. Once this limit is reached, the app will stop restarting and exit permanently. This change improves app stability by allowing transient issues to be recovered automatically while protecting users from repeated failures. It will log to both std error and log file, previously implemented in v1.4.0

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -2574,7 +2574,7 @@ fn setup_logging(app_handle: &tauri::AppHandle) {
     );
 }
 
-fn control_main() {
+fn main() {
     tauri::Builder::default()
         .plugin(tauri_plugin_os::init())
         .plugin(tauri_plugin_fs::init())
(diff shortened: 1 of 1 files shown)


Riccardo-Adami (approved):

CyberTimon:
Hey @riccio8, thanks for the PR.

This currently functions more as a crash restarter than a prevention tool. In my experience, crashes are rarely random; they're almost always deterministic issues related to environment variables or system configs, which a simple restart won't fix.
Could we adjust this PR to be more of a self-healing mechanism? For example, when a crash is detected, we could try applying some known environment variable fixes before attempting to restart. That would address the root cause of the most common issues.
What do you think of that approach?

riccio8:
> Hey @riccio8, thanks for the PR.
> 
> This currently functions more as a crash restarter than a prevention tool. In my experience, crashes are rarely random; they're almost always deterministic issues related to environment variables or system configs, which a simple restart won't fix. Could we adjust this PR to be more of a self-healing mechanism? For example, when a crash is detected, we could try applying some known environment variable fixes before attempting to restart. That would address the root cause of the most common issues. What do you think of that approach?

Yeah It's better, if you want I can implement a watch dog, this way, RapidRAW would become a child process of the watchdog, so that if it crashes the first time, it reloads it by changing variables from the environment such as the GPU backend or other things. If, even after 5 crashes and changes, it continues to cause problems, it simply shuts down. What do you think @CyberTimon?

CyberTimon:
Based on our discussion on Discord & [other GitHub issues](https://github.com/CyberTimon/RapidRAW/issues/418#issuecomment-3467535155), let's close this.

See: https://github.com/CyberTimon/RapidRAW/commit/390261bbe5a2c40669eb19665843c50043dbd477

Issue #40, opened 2025-07-09
No EXIF data found in this file.


belst:
When looking at metadata, it says there is no EXIF data, even though the file contains metadata.

a13-studios:
I just found out about this from Reddit. Was working on something similar - I can add some Exif support

CyberTimon:
EXIF metadata support should already be there. What kind of file are you using? 

This is the Metadata Panel on a DJI Air 3's RAW dng's:

<img width="771" height="1079" alt="Image" src="https://github.com/user-attachments/assets/96e28e59-4697-4ae6-8a9f-ecb17116670a" />

belst:
Fuji X-Trans Raw file (.RAF)

exiftool shows it correctly

belst:
example file for testing: https://bel.st/DSCF0717.RAF

CyberTimon:
Thank you. Will look into this in the coming days.

saltpit:
Same problem appears with photos from Panasonic DMZ-FZ200. 

Here's an example file: https://www.transfernow.net/dl/202509189rBmmxag

CyberTimon:
Thank you for your info. Sadly it's not something I can prioritize right now, as support for specific camera type's metadata depends on support of the underlying rawloader library: https://github.com/dnglab/dnglab/issues

saltpit:
Thanks for your answer. I will post my problem there.

riccio8:
> Thank you for your info. Sadly it's not something I can prioritize right now, as support for specific camera type's metadata depends on support of the underlying rawloader library: https://github.com/dnglab/dnglab/issues

Can this be closed?

CyberTimon:
Yes - let's close this issue, as it belongs to another library.

belst:
Even though the issue might be in an external library, it is still an issue with RapidRAW because it does not work there (regardless where the actual issue lies), so imo this should be kept open, and maybe reference a related issue in the dnglab repo until it is resolved

savukadoo:
I have also encountered this issue with the Fujifilm X-T5 RAF files, no EXIF data. I do agree that it should remain open as @belst mentioned as RapidRAW is leveraging off this library.

I will raise it on the link you sent if its not there already as this camera is 3 years old already and is very popular.

Cheers

CyberTimon:
[...]

PR #450, opened 2025-11-13
fix: use rawler to decode exif

sergiusens:
Serves at least as a temporary measure.

I tested with what I have ORF and RAF. I used claude to write the boilerplate (having already partially done this manually for another project)

image

image

Alleviates #40 

Signed-off-by: Sergio Schvezov <sergio@schvezov.net>

### DIFF
--- a/src-tauri/src/file_management.rs
+++ b/src-tauri/src/file_management.rs
@@ -20,6 +20,8 @@ use image::{DynamicImage, GenericImageView, ImageBuffer, Luma};
 use little_exif::exif_tag::ExifTag;
 use little_exif::metadata::Metadata;
 use num_cpus;
+use rawler::rawsource::RawSource;
+use rawler::decoders::RawDecodeParams;
 use rayon::ThreadPoolBuilder;
 use rayon::prelude::*;
 use regex::Regex;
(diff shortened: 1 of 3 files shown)


CyberTimon:
Thank you very much for your work. It’s great to see pull requests moving in this direction.

I noticed that you are now formatting the EXIF data in the backend. With the previous EXIF library, all EXIF data was extracted directly from the image and then sent to the frontend. The MetadataPanel is responsible for mapping the programmatic names to the displayed values, including the “All EXIF Data” section.

Have you already tried simply fetching the raw EXIF data from the backend and handling the parsing in the frontend? The current approach introduces an issue: part of the parsing happens in the backend, and then parsing happens again in the frontend. I hope my question makes sense.

Thank you again for your work.

sergiusens:
I understand, I only added the parsing later on as I was seeing raw exif data, but I will take another look and if I can figure out why what was happening

sergiusens:
I found some time @CyberTimon, and have pushed an update

sergiusens:
rebased to remove conflicts

CyberTimon:
Implemented in https://github.com/CyberTimon/RapidRAW/commit/d9a97a30c7f14c77b2c32ea84d773eb00270f1c7. Thanks for the PR, which guided some parts of the work.

PR #451, opened 2025-11-14
Implement new raw/jpeg grouping

EnTeQuAk:
I'm using a Fuji camera and regularly shoot JPEG+RAW to have the best of both worlds: straight out of camera JPEGs which are 80% of the time great, and RAWs for special occasions. Seeing every shot twice in the library has always bugged me, so I started experimenting with grouping.

The core idea: Rust assigns a `group_id` to images that share a stem in the same directory (e.g. `DSCF0644.RAF` and `DSCF0644.JPG`), and the frontend collapses each group into a single entry, picking a primary based on your preference. Virtual copies inherit the group but don't count toward it, so a file + its VC alone won't create a false group.

What I implemented:

- "Group Variants" in the file type filter, with a RAW/JPEG preference toggle underneath
- A small layers icon on grouped thumbnails; hovering it shows the extensions (e.g. "RAF+JPG")
- Variant switcher pills in the editor toolbar with actual file extensions, so you see `[RAF] [JPG]` instead of generic labels
- Filmstrip stays on the primary when you switch to a non-primary variant, so the position doesn't jump around

Filter dropdown with Group Variants option and RAW/JPEG preference

Editor view with variant switcher pills

Along the way I also fixed a pre-existing bug where `delete_files_with_associated` would grab the wrong stem for filenames with dots (was using `split('.').next()`, now uses `Path::file_stem()`), and simplified the `hasAssociatedFiles` context menu check from O(n*m) string matching to a direct `group_id` lookup.

Group-aware operations (delete/copy/move/export) aren't in this PR yet. When you delete a grouped file, should it delete all variants? Offer a choice? I'd rather discuss that and build it as a follow-up than guess wrong here.


CyberTimon:
Thank you! This is very nice and I can see some use in this. Can you try to match the design of the other components in RapidRAW? (Padding, size etc).

image

Thank you a lot!

As for the associated detection for the delete operations: it currently just checks if the base filename (removing extensions) is identical.

EnTeQuAk:
Awesome, thanks a lot for your reply. Yeah, the design is far from final. I first need to play around with it more to ensure it's actually useful the way I built it :see_no_evil: 

> As for the associated detection for the delete operations: it currently just checks if the base filename (removing extensions) is identical.

Indeed, I'm currently investigating if it makes sense to make that a bit more explicit and more widely known in other parts of the application.

Right now I'm trying to figure out how to generalize different types of files, e.g, other cameras may have a "XXX + RAW" mode, not sure if JPEG is kinda like the absolute default in the industry.

As I said, it's experimental and will likely take a while to finish. Your feedback is much appreciated :)

da-anda:
subscribing to this PR since I am very much interested in this (I also shoot in jpeg+raw). This is one thing I am missing after I moved away from Lightroom

EnTeQuAk:
Alrighty, both of your replies motivated me to pick up the work sooner rather than later :grin: 

Now, I just pushed a complete rewrite from the original experiment. Given the old branch was 612 commits behind `main` I reset it and started fresh.

Key differences from the earlier iteration:

- I moved the grouping entirely to the Rust side of things. The old approach did stem-matching in JS at render time. Now `assign_group_ids()` runs once during the directory scan, and each `ImageFile` carries a new `group_id` as well as `is_raw` from the backend. The frontend just buckets/groups by `group_id`
- replaced `RawOverNonRaw` entirely. The old "Prefer RAW" filter and the new grouping feature were solving the same problem at different levels. I merged them now into one "Group Variants" filter option. Old saved settings with `rawOverNonRaw` are migrated automatically.
- The variant switcher uses real extensions now. Instead of generic "RAW" / "JPEG" labels, the editor pills show the actual extension (RAF, DNG, ARW, CR3, JPG, etc.). The preference toggle in the filter dropdown uses the generic RAW/JPEG labels since those are categories, and a folder can have multiple extensions mixed. 
 - Fixed a pre-existing bug in `delete_files_with_associated` where `split('.').next()` would grab the wrong stem for filenames with dots. Switched to `Path::file_stem()` with proper sidecar pattern matching.
 - Simplified `hasAssociatedFiles` in the context menu. It was doing `O(n*m)` `startsWith` calls matching across the full image list. Now it's a direct lookup against the pre-computed group variant counts, which should be much faster.
 
This pull request still needs group-aware operations (delete/copy/move with confirmation modals), though, I'd like to open that question rather than implement too much right now. What do you think operations like delete/copy/move should behave with the raw/jpeg grouping enabled? I'd build that as a follow-up in a separate pull request, I'd like to focus on the foundation here.

Screenshot_20260316_173303
Screenshot_20260316_173243

@CyberTimon, regarding your earlier feedback: the design now follows the existing component patterns (same padding, border-radius, and color tokens as the rest of the dropdowns and toolbars). The associated files detection for delete uses `group_id` from the Rust scan rather than the old JS stem-matching, which covers your point about checking base filenames.

Let me know what you think. Happy to hear your feedback on this :angel:

kevinschweikert:
> I haven't seen that in other software yet

Photomator has this kind of feature and groups the JPG and RAW file:

image

EnTeQuAk:
>  Photomator has this kind of feature and groups the JPG and RAW file:

I stand corrected :angel: but to my excuse, I do not own any Apple hardware :see_no_evil: (aside from the ipad that my kids are using).

I kind of like their more descriptive text on the picture more than my numbered grouping, which is quite ambiguous

image

Do you have the chance to dig deeper on how they're handling it throughout the software, in case you like how they built it?

EnTeQuAk:
I experimented with that idea, but particularly when viewing lots of pictures and scaling the gallery down, it quickly becomes "only text" (next to the image name).

I added an icon to indicate that this picture is grouped, which felt like the best of both worlds :thinking: 

image

agirilovich:
This feauture is essential for the migration from other DAM software.
The workflow: "export Lightroom catalog as a collection of RAWs + processed JPEGs", - is described in many other applications documentation. Primary in "How to switch from Lightroom" article.
So, one more vote for the request.

However in this case, having "Prefer JPEG" option is the crucial. Since, imported RAWs are just archived RAWs without any settings and JPEGs should be displayed on a main screen as a result of your job in the past.

I love how this option is implemented in the Zoner Studio, with few more (optional) conditions:
Zoner_grouping
The setting is available "globally" and "per folder". They allow/block grouping if timedate in EXIF is different. And by default, edited JPEGs are not grouped with RAWs.

Flohhhhh:
Cool! I've always wondered why Lightroom didn't have something like this. Really nice, and opens up a lot of possibilities I think.

EnTeQuAk:
thanks for that feedback @agirilovich, I started to build more around that kind of usage-pattern and added two new grouping options inspired by Zoner Studio's approach.

"Group edited files" checkbox: When unchecked, files that have been independently edited in RapidRAW are excluded from grouping and appear as standalone images. This is particularly, I think, useful for Lightroom migrants where exported JPEGs can represent finished work and shouldn't be collapsed behind the archived RAW.

"Require matching metadata" checkbox: When checked, files sharing a filename stem are only grouped if their EXIF creation timestamps match. That should prevent false groupings when unrelated files from different cameras/shoots happen to share a filename. Files without EXIF data are excluded from grouping entirely when this is enabled.

Both options appear under the Group Variants filter in the View Options dropdown, alongside the existing RAW/JPEG preference toggle. Defaults preserve current behavior (group edited files: on, require matching metadata: off).

Not yet implemented (I think I'd do that in a separate PR): Per-folder grouping settings to allow each folder to override the global grouping preferences independently (similar to Zoner Studio's "Save settings for each folder separately" option). This would let users have different grouping behavior for e.g. their Lightroom import folder vs. their native RAW shooting folders. But, that's a lot of work from what I'm seeing and I don't wanna unnecessarily increase the size of the PR, it's hard to maintain as it is already.

What do you think? I'm also still testing it a bit more, so, expect dragons here and there.

agirilovich:
Right, you understand both options in the same way as Zoner Studio has it implemented. 


Few additional thinks:

> "Group edited files" checkbox: ..... where exported JPEGs can represent finished work and shouldn't be collapsed behind the archived RAW.

Do you have any way to recognize if JPEG file directly from camera or edited/exported?
It might be great feature, but the effort/usability ratio looks quite low.

I do belive, that at this stage of project development it is fully enough just having 3x options on a folder level:

- Group nothing.
- Group automatically, JPEG behind RAW
- Group automatically, JPEG over RAW. 



> "Require matching metadata" checkbox: When checked, files sharing a filename stem are only grouped if their EXIF creation timestamps match. That should prevent false groupings when unrelated files from different cameras/shoots happen to share a filename. Files without EXIF data are excluded from grouping entirely when this is enabled.

Exactly!  This option is more usable on upper level, when hundreds of files from different folders (recursively) are displayed on the same view. And even more in search views.

 
> Both options appear under the Group Variants filter in the View Options dropdown, alongside the existing RAW/JPEG preference toggle. Defaults preserve current behavior (group edited files: on, require matching metadata: off).

Hm. I've just compiled RapidRaw from a master branch and didn't find the option to display JPEG over RAW. Will check later....


> This would let users have different grouping behavior for e.g. their Lightroom import folder vs. their native RAW shooting folders.

It would be great if you were able to implement it one day. Great step toward building really mature and professional product.

EnTeQuAk:
> Do you have any way to recognize if JPEG file directly from camera or edited/exported?
> It might be great feature, but the effort/usability ratio looks quite low.

I don't think we do. But maybe @CyberTimon knows something I don't :angel: 

> Hm. I've just compiled RapidRaw from a master branch and didn't find the option to display JPEG over RAW. Will check later....

The changes in this branch are not on `main` yet

dedeweb:
That is a feature i'm waiting for, I hope it will be merged soon :crossed_fingers: :heart_eyes:

EnTeQuAk:
@CyberTimon alrighty, I think this branch is ready for review. You're moving incredibly fast, which makes it hard to keep up with `main`, but I've rebased (rather rebuilt) the feature on top of the latest `main` (~270 commits since the original PR, including the sx split, MainLibrary refactor, Zustand migration, etc.).

## What's in this branch

**Backend** - `is_raw` and `group_id` fields on `ImageFile`, `assign_group_ids()` that buckets files by filename stem with optional EXIF date verification via `try_get_exif_creation_date()`. Runs consistently across `list_images_in_dir`, `list_images_recursive`, and `get_album_images`.

**Frontend** - `buildImageGroups()` collapses groups in the display list. Grouping UI in the view options dropdown (Off / Prefer RAW / Prefer JPEG) with sub-options for metadata matching and edited file handling. The `RawOnly`/`NonRawOnly` filter now uses the backend-provided `image.is_raw` instead of recomputing from extension lists. Legacy migration from the old `rawOverNonRaw`/`groupVariants` settings.

**Polish** - Group badges on thumbnails (Layers icon with tooltip like "JPG+RAF"), variant switcher pills in the editor toolbar (segmented control style), filmstrip highlight sync so the correct thumbnail stays selected when switching variants. Hidden variants are still reachable through search and filter. "Delete with associated files" now uses `group_id` instead of filename prefix matching.

image

image

image

## Architecture differences from the original branch

- Grouping is a first-class setting (`appSettings.grouping: 'off' | 'raw' | 'jpeg'`) rather than overloading the raw status filter enum
- State lives in Zustand stores (great move on your side!)
- Utility functions (`buildGroupBadgeInfo`, `findGroupVariants`, `getVariantLabel`) are extracted into `imageGrouping.ts` rather than inline in App.tsx
- Explicit legacy migration in `useAppInitialization.ts` for old settings
- All grouping labels are i18n-ready via `useTranslation`

Let me know if I should prepare the branch in any other way or if you need more context. I'm still testing, but getting more eyes on it at this point should be helpful.

EnTeQuAk:
A few more follow-up fixes:

**Search vs filter semantics** (`useSortedLibrary.ts`)
- Filters (rating, raw status, edited status) now apply to the group primary only. Previously, the group-aware matching applied to both filters and search, which caused false positives: e.g. "Edited Only" filter would surface an unedited RAW because its hidden JPEG was edited.
- Search (text/tags) remains group-aware: if any variant matches the query, the group primary surfaces. This is the right behavior since the user is looking for something specific by name or tag.

I did some smaller type-safety refactoring and added a `ViewOptionsDropdownProps` interface with proper types for all 14 props (was `any`). Uses `ThumbnailSize`, `ThumbnailAspectRatio`, `RawStatus`, `EditedStatus` enums instead of bare strings.
- Removed all `: any` annotations from `.map()` callbacks that are now properly typed through the interface.
- Fixed `setFilterCriteria` callback signatures: `prev` was annotated as `Partial<FilterCriteria>` but the store passes a full `FilterCriteria`. This was a pre-existing error in the file, but I fixed it since I already touched that file anyway..

I updated labels to:
- "Group even excluded items" → "Include edited files in groups"
- "Group even with differing metadata" → "Ignore mismatched EXIF dates"

I hope that makes things clearer

ruralmeltdown:
@EnTeQuAk amazing work on this, hopefully it gets merged in soon. Quick question about it that I might have missed in your updates above: in Lightroom, you usually see the jpg as the thumbnail, but then see the raw when you go to edit. Will your changes act like this same way, or something different?

EnTeQuAk:
> @EnTeQuAk amazing work on this, hopefully it gets merged in soon. Quick question about it that I might have missed in your updates above: in Lightroom, you usually see the jpg as the thumbnail, but then see the raw when you go to edit. Will your changes act like this same way, or something different?

Not really no, I'm sorry.  If I remember right, Lightroom's behavior is "JPEG thumbnail, RAW on edit". Our current implementation is almost opposite of that:

- Prefer RAW = library shows the RAW thumbnail, double-clicking opens the RAW in the editor.
- Prefer JPEG = library shows the JPEG thumbnail, double-clicking opens the JPEG in the editor.

The variant pills in the editor toolbar let you switch between the two once you are already editing, but the initial editor load is whichever variant you set as the primary.

ruralmeltdown:
> Not really no, I'm sorry.  If I remember right, Lightroom's behavior is "JPEG thumbnail, RAW on edit". Our current implementation is almost opposite of that:
> 
> - Prefer RAW = library shows the RAW thumbnail, double-clicking opens the RAW in the editor.
> - Prefer JPEG = library shows the JPEG thumbnail, double-clicking opens the JPEG in the editor.
> 
> The variant pills in the editor toolbar let you switch between the two once you are already editing, but the initial editor load is whichever variant you set as the primary.

Oh, okay, thank you for the clarification. I was reading this as more of a Lightroom-like implementation of that, but I think having the ability to see JPEGs in the library, then go to edit it, and be able to toggle between the RAW and JPEG might actually be a better implementation (I don't even think LR does that, you just see the RAW on the Develop tab). I look forward to this being rolled out.

CyberTimon:
Hi @EnTeQuAk,

Thank you so much for your continued hard work on this!

I've tested it thoroughly and I’m really happy with how it turned out. The performance, UI, and overall implementation look fantastic.

I’m going to merge this now, and I’ll do a quick follow-up commit to touch up a few code comments.

Have a nice weekend!
Timon

EnTeQuAk:
Woohooo, thanks @CyberTimon for merging and the follow-up cleanups, really appreciated.

PR #462, opened 2025-11-20
Migrate to React-Window v2

RayJW:
This is my 2nd attempt at getting this done. I was able to remove autosizer and follow the new recommended way of doing things according to the `react-window` guide. At least in my testing on Linux, all the functionality is the same while getting rid of dependencies and finally adopting the newer API for react-window. As always, please test if these changes also work for you.

Some sizing logic had to be changed, but I made sure that it's still reactive while making things faster (from my limited comparison of render times). I left some comments for a few things I stumbled on while developing this, but if it's preferred, I can remove them before merging.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1479,9 +1479,9 @@
       ]
     },
     "node_modules/@tauri-apps/api": {
-      "version": "2.9.0",
-      "resolved": "https://registry.npmjs.org/@tauri-apps/api/-/api-2.9.0.tgz",
-      "integrity": "sha512-qD5tMjh7utwBk9/5PrTA/aGr3i5QaJ/Mlt7p8NilQ45WgbifUNPyKWsA63iQ8YfQq6R8ajMapU+/Q8nMcPRLNw==",
+      "version": "2.8.0",
+      "resolved": "https://registry.npmjs.org/@tauri-apps/api/-/api-2.8.0.tgz",
+      "integrity": "sha512-ga7zdhbS2GXOMTIZRT0mYjKJtR9fivsXzsyq5U3vjDL0s6DTMwYRm0UHNjzTY5dh4+LSC68Sm/7WEiimbQNYlw==",
       "license": "Apache-2.0 OR MIT",
       "funding": {
         "type": "opencollective",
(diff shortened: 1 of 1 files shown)


CyberTimon:
Thank you - it works correctly on my system.

Regarding the code quality: I noticed that several fallback paths were included to infer which new APIs react-window v2 might use. Now that the working APIs are known, those conditional branches can be removed. Please also remove the code comments.

Thanks again.

RayJW:
> Now that the working APIs are known, those conditional branches can be removed.

Yes, sorry, that actually reminds me that there is still a regression when navigating back the scroll position isn't being restored. I'll have to fix that before being ready.

Shouldn't finish code like this late at night, sorry ^^'

CyberTimon:
No problem. Thanks for the PR :)

CyberTimon:
Hi @RayJW

While implementing https://github.com/CyberTimon/RapidRAW/issues/466, I also migrated to react-window v2. Everything is running fine now, except for two things:

- In react-window v2, the scroll height is accessed like this: `onScroll={(event) => setLibraryScrollTop(event.currentTarget.scrollTop)}`. Without a debounce this causes noticeable lag.
- I burned way too many hours today trying to restore the scroll position after leaving the editor. I have no idea why, but it genuinely feels impossible in v2 ._.

I’m sorry that I ended up doing most of the work and couldn’t merge yours. Re-basing after the #466 changes would have been a nightmare.

If you’re up for it, I would **really appreciate help with restoring the scroll offset/position after leaving the editor**.

Thank you!

RayJW:
> * I burned way too many hours today trying to restore the scroll position after leaving the editor. I have no idea why, but it genuinely feels impossible in v2 ._.

I feel you, this has been most of the work for me as well.

> I’m sorry that I ended up doing most of the work and couldn’t merge yours.

Dw about it, I'm happy this can be moved along any way :)

> If you’re up for it, I would **really appreciate help with restoring the scroll offset/position after leaving the editor**.

From what I gathered in the docs, the best possible way would be storing the number of the first row in view and then restoring that with the `scrollToRow` property and `behaviour: "instant"`. But I couldn't quite make that work yet in an elegant way.

CyberTimon:
Reverted back in the meantime as it really looks like a direct `scrollTo` is not possible with React-Window V2. Maybe we can take a look at this again in the future, or if you want, @RayJW, feel free to keep trying 💯 

Commit: https://github.com/CyberTimon/RapidRAW/commit/b1d5aee0dca5155aa3fccfc1aafbeed7087807d1

RayJW:
I'll close this, as too much has changed in the meantime for this to still make sense without starting over.

PR #463, opened 2025-11-20
Fix Warnings on Linux

RayJW:
This makes the theme and window variables that are exclusive to Windows and macOS as intentionally unused to get rid of the build warnings on Linux.


CyberTimon:
Thanks - LGTM.

PR #475, opened 2025-11-27
mem optimizations - reduce mem usage 

william-sy:
- Added max images in mem setting
- Memoized the right-hand adjustments panel with a custom comparator so it only re-renders when its key props (adjustments, selected image, collapsible state, etc.) actually change
- updated packages

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -400,33 +400,19 @@ function App() {
   const imagePathList = useMemo(() => imageList.map((f: ImageFile) => f.path), [imageList]);
   const [thumbnails, setThumbnailsState] = useState<Record<string, string>>({});
   const thumbnailCacheRef = useRef<Map<string, string>>(new Map());
-  const [thumbnailMemoryLimit, setThumbnailMemoryLimit] = useState<number>(300);
-
-  const enforceThumbnailLimit = useCallback(
-    (limitOverride?: number) => {
-      const cache = thumbnailCacheRef.current;
-      const effectiveLimit = Math.max(
-        50,
-        Math.min(limitOverride ?? thumbnailMemoryLimit ?? 300, 2000),
-      );
-      while (cache.size > effectiveLimit) {
-        const oldestKey = cache.keys().next().value;
-        if (oldestKey === undefined) break;
-        cache.delete(oldestKey);
-      }
-      setThumbnailsState(Object.fromEntries(cache));
-    },
-    [thumbnailMemoryLimit],
-  );
 
   const resetThumbnails = useCallback(
-    (next: Record<string, string>) => {
+    (next: Record<string, string> | ((prev: Record<string, string>) => Record<string, string>)) => {
       const cache = thumbnailCacheRef.current;
+      
+      // Handle both direct value and updater function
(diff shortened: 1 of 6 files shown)


william-sy:
@CyberTimon, You want this pr to remain open or shall i close it ?

CyberTimon:
I will review it in the coming days. I haven't had time (/ priority) this weekend.

william-sy:
No problem, thanks for the update much appreciated.

CyberTimon:
Hi @william-sy 

I finally found some time to properly test and evaluate this PR :)

Here are my thoughts:
 - Undo History: These entries aren't very memory heavy. I worry that a max history limiter will cause more issues than it solves.
 - LUT Cache: Since LUTs are only ~4-5MB and rarely used, adding a cache here feels like it complicates the code for very little return.
 - Thumbnails: A limit is good, but the fixed count is tricky. If the limit is 50 and a folder has 150 images, the user experience breaks. Maybe we could just flush the cache on folder switch? This seems like the best area to target for optimization.
 
Did you have any specific data regarding the first two points? (i.e., did you see a major RAM drop with the history/LUT changes?).
 
Don't get me wrong, I really appreciate the effort here! I just want to avoid overengineering the whole codebase to save a few MBs when the main RAM leak might be hiding somewhere else ;)
 
Looking forward to your feedback!
 
Timon

Edit:
**One additional point I noticed regarding the React optimizations:**

In `src/components/panel/Editor.tsx` and `src/components/panel/right/ControlsPanel.tsx`, you’ve implemented `React.memo` using a custom `arePropsEqual` comparator that manually checks a long list of props.

```javascript
function arePropsEqual(prev: EditorProps, next: EditorProps) {
  return (
    prev.selectedImage?.path === next.selectedImage?.path &&
    prev.finalPreviewUrl === next.finalPreviewUrl &&
    // ... extensive list of props
  );
}
```

While I understand the intention is to prevent unnecessary re-renders, this pattern introduces significant **maintenance debt**. If we add a new prop to these components in the future (e.g., a new view mode or toggle) but forget to manually add it to this return statement, the UI will silently fail to update.

william-sy:
@CyberTimon 

Thank you for the review. Currently with your suggestions I still get a 75% mem reduction. 
So the following changes have been applied: 
- Remove history limiter
- Flush thumbnail cache on folder switch 
- Remove custom memo comparators 
- Remove thumbnail memory limit UI setting

CyberTimon:
Thanks for the update! However, I notice this PR still includes the LUT caching logic in the Rust backend (MAX_LUT_CACHE, LutCacheEntry, etc.) which I mentioned previously adds complexity without enough benefit.

Additionally, the changes to App.tsx regarding thumbnailCacheRef and the backend return types change the architecture significantly more than I am comfortable with for a bug fix.

Since the primary memory fix is likely just flushing the thumbnails on folder change, I'm going to close this PR. I will implement a clean patch that simply resets the thumbnails state when the folder changes. Thank you for investigating where the leak was coming from!

Have a nice day :)
Timon

CyberTimon:
Thumbnail cache flush on folder change is now implemented in https://github.com/CyberTimon/RapidRAW/commit/619d9eabf30e6a59f0ef494127c8ace695f0f43d. 💯

PR #513, opened 2025-12-10
Removed dashed line for Brush Mask

Duecki1:
This commit removes the dash for the brush mask and its eraser.

Due to the new eraser the dashed lines get confusing very fast, so I removed them. I hope that is fine.

before: 

image

After:

image


CyberTimon:
Thank you for the PR! Looks good to me - I also like it this way. 💯

PR #519, opened 2025-12-17
Merging to hdr image

SimonIT:
It's not quite there yet, but please feel free to provide some initial feedback.

- [x] Getting exposure and gain from the raw
- [x] `to_rgb8()` for preview(?)
- [x] Saving

### DIFF
--- a/src-tauri/src/exif_processing.rs
+++ b/src-tauri/src/exif_processing.rs
@@ -5,7 +5,7 @@ use std::path::Path;
 
 use crate::formats::is_raw_file;
 use chrono::{DateTime, Utc};
-use exif::{Exif, Field, In};
+use exif::{Exif, Field, In, Value};
 use little_exif::exif_tag::ExifTag;
 use little_exif::filetype::FileExtension;
 use little_exif::metadata::Metadata;
@@ -70,23 +70,35 @@ pub fn read_exposure_time_secs(path: &str, file_bytes: &[u8]) -> Option<f32> {
 
     if let Some(exif) = read_exif(file_bytes) {
         if let Some(exposure) = exif.get_field(exif::Tag::ExposureTime, In::PRIMARY) {
-            let num = exposure.value.get_uint(0)?;
-            let denom = exposure.value.get_uint(0)?;
-            return if denom == 0 {
-                None
-            } else {
-                Some(num as f32 / denom as f32)
-            };
+            if let Value::Rational(ref r) = exposure.value {
+                if r.is_empty() {
+                    return None;
+                }
+
+                let val = r.get(0)?;
+
+                return if val.denom == 0 {
+                    None
+                } else {
+                    Some(val.num as f32 / val.denom as f32)
+                };
+            }
         } else if let Some(shutter_speed) =
             exif.get_field(exif::Tag::ShutterSpeedValue, In::PRIMARY)
         {
-            let num = shutter_speed.value.get_uint(0)?;
-            let denom = shutter_speed.value.get_uint(0)?;
-            return if denom == 0 {
-                None
-            } else {
-                Some(num as f32 / denom as f32)
-            };
+            if let Value::Rational(ref r) = shutter_speed.value {
+                if r.is_empty() {
+                    return None;
+                }
+
+                let val = r.get(0)?;
+
+                return if val.denom == 0 {
+                    None
+                } else {
+                    Some(val.num as f32 / val.denom as f32)
+                };
+            }
         }
     }
     None


SimonIT:
Here's a sneak peek:

https://github.com/user-attachments/assets/f99391a4-ab85-4f38-a930-ec53a39e4eea

I have already replaced 'panorama' with 'HDR' in the texts.

The reading of the EXIF values may need improvement

CyberTimon:
Thanks @SimonIT for this PR. I will take a look at this soon. Do you already have a fix in mind for these purple sensor clipping artifacts?

SimonIT:
Thanks! Perhaps it's because of the hard-coded highlight compression to 2.5? You used this value for culling, denoising and panorama stitching too, though, so I thought it would be OK. It may simply be how RapidRaw behaves with overexposed parts. When I reduce the exposure by -2 on the second mushroom picture (_1020394.RW2), the image just has purple clipping. I can try with different pictures, though — I wasn't planning to merge these when I took them.

CyberTimon:
Hi @SimonIT,

Thanks for the great work on this PR and for the demo video! I've had a chance to look through the code, and it's integrating nicely.

I have two main suggestions to align it with the rest of the project:

1.  **EXIF Data Reading**: I saw you use the `image-hdr` crate to read metadata. That crate uses an EXIF library that has limited support for RAW files, which can lead to incorrect exposure or ISO values. Instead, please use RapidRAW's own functions from `exif_processing.rs`. The `read_exif_data` function is perfect for this, as it's built on `rawler` and will give you the correct values from the source files. This will also help avoid duplicating code.

2.  **Logging**: The code currently uses `println!`. Could you please switch this to use the `log` crate macros (e.g., `log::info!`, `log::debug!`)? This ensures all logging is handled consistently by the application's logger.

After these changes, I'll do some final testing, and we should be ready to merge. 

Thank you & have a nice evening,
Timon

SimonIT:
I changed it to only use the RapidRaw Exif reading, removed and/or replaced the println, and made the error handling (hopefully) in general better for that function

CyberTimon:
Hey @SimonIT, thanks for the updates!

Two things:

**1. Replace `parse_exposure_time` with typed accessors**

The `parse_exposure_time` function needs to be replaced. It's stripping arbitrary characters and guessing at formats, which is defensive coding against a problem we shouldn't have. `read_exif_data` formats exposure as `"1/200 s"`, then `parse_exposure_time` reverse-engineers that string back into a number. It works by accident today, but it's two functions fighting each other through an untyped string. I'd rather not have this pattern in the codebase.. once we start guessing at our own internal formats, it tends to spread.

The rational values already exist in `extract_metadata`, they just need to be exposed directly. Could you add typed accessors to `exif_processing.rs`?

```rust
pub fn read_exposure_time_secs(path: &str, file_bytes: &[u8]) -> Option<f32> {
    // return num/denom from the Rational directly, no string involved
}

pub fn read_iso(path: &str, file_bytes: &[u8]) -> Option<f32> {
    // same
}
```

Then use these in `merge_hdr` and remove `parse_exposure_time` entirely. No formatting, no parsing, no guessing ;)

**2. Use `log::info!` instead of `debug!`**

The rest of the codebase uses `log::info!` for this kind of operational logging. Could you switch the `debug!` calls like `"Read image {} with gains: {}"` and `"Read image {} with exposure: {}"` to `log::info!` for consistency?

Thanks a lot!
Timon

SimonIT:
Just did _Use log::info! instead of debug!_ quickly, I'm on _Replace parse\_exposure\_time with typed accessors_ right know

CyberTimon:
No stress!

I’d suggest taking the time to integrate this cleanly with the rest of the codebase instead of rushing the last steps. A slightly slower, tidy integration now will make maintenance much easier later :)

SimonIT:
It's done. The problem with waiting is that merge conflicts appear all the time. And I'm not patient^^

CyberTimon:
Hi @SimonIT,

These changes look great! Thank you for carefully addressing the feedback and cleaning up the EXIF handling. I really appreciate the effort you put into this.

I’ll go ahead and merge it now. 💯

SimonIT:
🎉 Reading the meta data for raw should also be a bit faster, as it reads it now from the slice.

I love to contribute, but I haven't found the right topics as I don't have much knowledge in this field. I'm looking forward to see this awesome succeed and helping it.

By the way, the author of the image-hdr crate has also some other interesting crates like star detection and something other I don't really understand for denoising? Might be worth a look

CyberTimon:
> 🎉 Reading the meta data for raw should also be a bit faster, as it reads it now from the slice.

I saw this while reviewing the code - that's amazing!! thanks for this fix.

> I love to contribute, but I haven't found the right topics as I don't have much knowledge in this field. I'm looking forward to see this awesome succeed and helping it.

Thanks a lot - that's great and I really appreciate it 🥇 

> By the way, the author of the image-hdr crate has also some other interesting crates like star detection and something other I don't really understand for denoising? Might be worth a look

Yesterday I also looked at his other repositories. Yes, some of them are for noise reduction or feature detection (perhaps useful?). I don't currently have a direct use for them.

PR #525, opened 2025-12-21
Fix subject mask stop tracking when outside the image

Duecki1:
this commit lets the subject lasso continue outside of the actual image, instead of just canceling the lasso.

Before: 

https://github.com/user-attachments/assets/307eea5a-3e02-4ab0-a04c-3ebcbc12b949

After:

https://github.com/user-attachments/assets/6760b499-5e86-45da-9a4f-ef23105ad33d


CyberTimon:
Great! It works much better now. Thank you so much for the contribution!

Duecki1:
Glad I could help, I really love this project :)

PR #531, opened 2025-12-23
Bump All Dependencies

RayJW:
NPM was throwing warnings about unfixed vulnerabilities once again so I took that as a prompt to go through all the dependencies again to clean things up.

Tested it all extensively on Linux and couldn't find any regressions.


CyberTimon:
Works fine on my end too! Thank you for the PR. Tested on macOS and Windows.

Issue #538, opened 2025-12-25
Regression in v1.4.7: AI selection sometimes follows the mouse with a huge diagonal offset


patrakov:
### What happened?

I tried to add an AI subject selection while zoomed in, on a less-powerful laptop (with only 16 GB of RAM).

The first selection went well. The second selection, where I encircled the fragment of a building wall, was not following my mouse pointer. Instead, it painted blue lines between the actual pointer position and something diagonally offset.

See the video below.

https://github.com/user-attachments/assets/b9af8de4-0041-4101-bf6f-c75f7434b94a

### What did you expect to happen?

A blue line following the mouse pointer while I was moving it.

### How can we reproduce this?

On Windows 11:
 
1. Open a RAW file.
2. Create a mask with an AI subject.
3. Select something.
4. Zoom in - that's sufficient to make rapidRAW sluggish.
5. Add another AI subject mask.
6. Try to select something else.

### Operating System

Windows 11

### RapidRAW Version

v1.4.7

### Additional information

The bug does not exist on v1.4.6.

Note that the issue is with providing the input to the AI model, not with the model itself. The stuff selected in blue is not what I encircled.

### Similar issues

- [x] I have searched existing issues and didn't find a duplicate

patrakov:
The slowness (caused by zooming in) was a red herring. The issue is perfectly reproducible on Windows with HiFi zoom off.

In the video below, the first two components of the AI subject mask behave as they should. The third one glitches out.

https://github.com/user-attachments/assets/01ae2354-f08d-47aa-bf33-e1b15dc7a686

patrakov:
Reproducible on Arch Linux, too (after manually bumping the version in the PKGBUILD). Took 12 tries.

Duecki1:
@patrakov just curious, does this also happen with high quality zoom turned off on the settings?

CyberTimon:
@Duecki1 Probably since we merged https://github.com/CyberTimon/RapidRAW/pull/525. Will take a look soon but would be cool if you could see if you can fix it.

CyberTimon:
I can also reproduce it (with or without high quality zoom). It seems to happen when you start drawing while the image itself is processing in the background. It happens more when using high quality zoom as the loading phase is longer then.

Duecki1:
> I can also reproduce it (with or without high quality zoom). It seems to happen when you start drawing while the image itself is processing in the background. It happens more when using high quality zoom as the loading phase is longer then.

[...]

PR #541, opened 2025-12-26
Fix the issue when ai masking under load

Duecki1:
This should fix the issue: https://github.com/CyberTimon/RapidRAW/issues/538

I've tested it on my system and it works.


CyberTimon:
Thank you a lot for the quick fix. I'm not able to reproduce the problem anymore 👍😃

PR #579, opened 2026-01-14
Add extra issue type

william-sy:
Hey Timon, 

Noticed a lot of FR in the issue list, it would be great if they get a auto title and a auto label
so we can filter on the issue list and be more productive :)


CyberTimon:
Great! Thank you a lot for the template. It looks good to me 💯

CyberTimon:
@william-sy

Wait, we now have 2 issue title fields:
image

william-sy:
Ah yes, I see, one moment :)

PR #582, opened 2026-01-15
Template fix: remove duplicate title

william-sy:


CyberTimon:
Thanks for the quick fix - LGTM

PR #593, opened 2026-01-21
Test Android Port

Duecki1:
This a test for an Android Port

The Tauri V2 Android Documentation is not that informative imo. I used Ai Studio for help.

For the first time run: 

npm install
npm run tauri android init

And then build: 
npm run tauri android build -- --debug --target aarch64

The app opens, uses Androids File Picker but still crashes on edit.

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -34,7 +34,7 @@ tokio = { version = "1", features = ["full"] }
 uuid = { version = "1.19", features = ["v4", "serde"] }
 walkdir = "2.5.0"
 imageproc = "0.25.0"
-ndarray = "0.16"
+ndarray = "0.17"
 reqwest = { version = "0.12", default-features = false, features = ["json", "multipart", "rustls-tls"] }
 tokio-tungstenite = { version = "0.28", features = ["rustls-tls-webpki-roots"] }
 futures-util = "0.3"
--- a/src-tauri/src/tagging.rs
+++ b/src-tauri/src/tagging.rs
@@ -16,10 +16,10 @@ use tokio::task::JoinHandle;
 use walkdir::WalkDir;
 
 use crate::AppState;
-use crate::candidates::TAG_CANDIDATES;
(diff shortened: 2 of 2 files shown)


Duecki1:
https://github.com/user-attachments/assets/3eceefcc-4fb4-42c0-a2ab-f2f24f4ad1df

william-sy:
The video is quite short, and the PR wow that is massive... but at 0:03 there are quite a few layout problems visible. It does look to be working but does not look functional... 

That being said though, If you could make and build a APK in your local fork and link us so we can try it. out that would be great!

Duecki1:
> The video is quite short, and the PR wow that is massive... but at 0:03 there are quite a few layout problems visible. It does look to be working but does not look functional... 
> 
> That being said though, If you could make and build a APK in your local fork and link us so we can try it. out that would be great!

Yes, I think this would be more fitting for tablets. For mobile phones, the UI will have to be optimized. 

The PR is massive due to many new generated files and the video is short because the app crashes instantly most of the time.

I'm still not sure about if this is really the best way or if a dedicated android application using compose or something is better.

The compose app will probably run better/be more integrated into Android, but since it's a sperate application, its much more work. This port using tauri will probably also need some changes in the raw processing code for it to run smoothly, which will make this repo more complex.

And the most annoying part for me the tauri v2 for android documentation is really not helpful. (And it really isn't the best way of developing for android)


The app: [Google Drive](https://drive.google.com/file/d/14VnKQ7ehHJMCMKVYWVzDmuOjwd_bILtl/view?usp=drivesdk)


(The app seems to use too much ram when moving the sliders, causing background tasks to stop working, for example music playback)

Duecki1:
Screenshot_20260304_210413_RapidRAW

S25Ultra
Refined some things with the storage system of android.

Duecki1:
https://github.com/user-attachments/assets/1082b478-5062-450f-9609-b3e67c9fe30a

Duecki1:
Will open a new one with better code soon

PR #668, opened 2026-02-13
fix: estimate batch export size with original image size

kevinschweikert:
In the batch export window for the same image I had some wildly different values than in the single image export mode

Batch Mode: ~230KB
Single Mode: ~5.8MB

Turns out, the boolean flag for fast de-mosaic generates a lower resolution image and this throws off the size calculations. I couldn't find out how to load the fast de-mosaic version in the preview size but still get the values for original size. Is there something exposed in rawler?

Before:

```
2026-02-13 20:47:10 [INFO] crop: Rect{6:5, 1548x1032, LTRB=[6, 5, 1554, 1037]}, intermediate dim: Dim2 { w: 1561, h: 1042 }, rawimage: Dim2 { w: 6656, h: 4608 }
```

After:

```
2026-02-13 20:43:55 [INFO] crop: Rect{26:20, 6192x4128, LTRB=[26, 20, 6218, 4148]}, intermediate dim: Dim2 { w: 6244, h: 4168 }, rawimage: Dim2 { w: 6656, h: 4608 }
```

The values are still not exactly the same but close enough

Batch Mode: ~6.1MB
Single Mode: ~5.8MB

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -1945,7 +1945,7 @@ async fn estimate_batch_export_size(
     const ESTIMATE_DIM: u32 = 1280;
 
     let original_image = match read_file_mapped(Path::new(&source_path_str)) {
-        Ok(mmap) => load_base_image_from_bytes(&mmap, &source_path_str, true, highlight_compression, None)
+        Ok(mmap) => load_base_image_from_bytes(&mmap, &source_path_str, false, highlight_compression, None)
             .map_err(|e| e.to_string())?,
         Err(e) => {
             log::warn!(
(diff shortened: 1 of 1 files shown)


CyberTimon:
Thanks for the PR @kevinschweikert. I noticed the same issue. 

The cleanest solution would be to scale the values back up so they match the correct resolution. The difficulty is that we don’t know whether the image is X-Trans (Fuji), which would be at 1/6 resolution, or Bayer, which would be at 1/4 resolution. I’m also not sure yet how to reliably detect this. Rendering the non-fast demosaiced version might be an option, but it will likely hurt performance (I haven’t tested it yet). What are your thoughts?

Timon

kevinschweikert:
I'm not too familiar with the code, but is there a way to get the two images sizes from the raw image? The logs suggest, that both information are available. We could then generate the preview with the smaller image and then scale that up to the final size for estimation. Then we shouldn't need to differentiate between X-Trans or Bayer because rawler handles that for us.

On the other hand, this will be calculated only for one image, so how much does it hurt performance to work with the full resolution?

CyberTimon:
Fixed in https://github.com/CyberTimon/RapidRAW/commit/e9eadbbc4a1041b861dd5ac19778481d84094f4a - I added a function to get the fast demosaic scale factor (in `raw_processing.rs`) and then used this to guess the actual image size.

Issue #694, opened 2026-02-16
FEATURE: export masks as layers


shsmad:
### Feature Description

I use not only raw-editor but also GIMP to polish image. I like quick masks and edits in rapidraw, but now if I want to edit mask borders more accurately I export image several times with masks edits applyed to whole image and make mask in GIMP.

It would be great if RapidRaw have functionality which would allow to export image not as single jpeg file but as collection of files, representing base image with adjustments and all masks as separate images with mask ajdustments and alpha-channel as mask border

### Similar feature requests

- [x] I have searched existing issues and discussions and didn't find a duplicate

CyberTimon:
This is implemented now! Will appear in the next release. Thanks for your contribution.

PR #696, opened 2026-02-16
Option to export masks.

shsmad:
If enabled, each mask is exported as an image with color effects for whole image + grayscale opacity mask.

Fixes https://github.com/CyberTimon/RapidRAW/issues/694

### DIFF
--- a/src-tauri/src/file_management.rs
+++ b/src-tauri/src/file_management.rs
@@ -197,6 +197,8 @@ pub struct ExportPreset {
     pub watermark_scale: u32,
     pub watermark_spacing: u32,
     pub watermark_opacity: u32,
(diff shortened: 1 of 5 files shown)


CyberTimon:
Hi @shsmad,

Thank you for this PR!

I've reviewed it and made a few improvements:
- Separate mask exports now work also when exporting multiple images
- When mask export is enabled, the base image no longer has mask adjustments applied (otherwise the feature wouldn't make much sense ;))
- Deduplicated the export resize functions and some other export related stuff in the backend while implementing this
- Integrated the new settings into export presets and other parts of the codebase

This looks great - merging now! 🥇

Thanks for the contribution!

Timon

PR #699, opened 2026-02-17
add refresh button to toolbar in Library section

savukadoo:
Adds a refresh icon to the Library section when new images are added and the user wants to refresh just that folder on the fly

### DIFF
--- a/src/components/panel/MainLibrary.tsx
+++ b/src/components/panel/MainLibrary.tsx
@@ -1166,6 +1166,14 @@ export default function MainLibrary({
     folder: null as string | null,
   });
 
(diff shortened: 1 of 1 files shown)


CyberTimon:
Thanks for the PR. 👍 

As I don't want to clutter the UI, adding a big button at the top for just doing this little task feels overkill.

I still don't get why users want such refresh button, as clicking on the current folder on the left does exactly the same:
image

Thanks for a quick explaination :)

savukadoo:
I see your point of view, I suspect my experience of using the app was that clicking the folder again was not clear in terms of it providng the refresh function. Im happy with that for now.

PR #704, opened 2026-02-18
Extra check for lens and camera support

william-sy:
Added 2 checkboxes.


kmilos:
Most of the recent requests were opened as features, not bugs...?

CyberTimon:
Improved some formatting, will merge now. Thanks!

william-sy:
Thanks both for feedback / improvement :D 
@kmilos, That is true, but by doing 1 template first we can just copy paste to FR later if needed

Issue #690, opened 2026-02-15
FEATURE: Sharpening mask


deneirgits:
### Feature Description

A new slider under Details > Sharpening to adjust the mask of where the sharpening would be applied. Increasing this slider would exclude non-edge parts of the elements in the image so sharpening would only apply on their edges.

Currently without this feature and with only the broad/imprecise masking options available, it's highly likely that sharpening would also be applied on areas where there's no detail or where sharpening is not desired, essentially just introducing noise.

### Additional Context

I'm a new RapidRAW user, trying it out from Darktable. If this is already possible with the current masking techniques, please do guide me 😄

Below are screenshots of this feature in Lightroom and Darktable

Lightroom:
<img width="1920" height="1040" alt="Image" src="https://github.com/user-attachments/assets/0f265e29-361c-4da2-a1cd-db4a686c1d28" />

Darktable:
<img width="1518" height="1736" alt="Image" src="https://github.com/user-attachments/assets/717a2c91-9409-442a-8a16-c0ea1668cf03" />

### Similar feature requests

- [x] I have searched existing issues and discussions and didn't find a duplicate

CyberTimon:
This is now implemented, altrough a visualization of the mask isn't here yet. I will close this for now as the core feature is in place. Will appear in some hours (when I release v1.5.5)

PR #705, opened 2026-02-18
feat: Sharpening Mask

d1anarene:
Lightroom Style sharpening mask slider
Applied area can be seen when dragging while pressing "Alt"

Requested in #690


CyberTimon:
Thanks for the PR. Every image now has a strange noisy overlay which I wasn't able to remove using sliders:

image

Do you have an idea what's going wrong? Is this PR really working for you?

Thanks!

d1anarene:
Will check again soon but yes it worked fine for me

CyberTimon:
Amazing - works pretty awesome! What do you think about showing the masking while dragging the slider and not just when holding down ALT? I'm unsure how many people will realize they have to press ALT.

What I noticed is that the threshold and the functionality in general is pretty inconsistent across RAW resolutions, preview resolutions and jpeg itself. Do you have a fix for this in mind?

Thanks a lot for your work!! :)

d1anarene:
Yeah you are right. I just thought I'll keep it consistent from LR, but I solely use that slider there while pressing alt.

I could try to normalize the sharpening threshold by image resolution so the mask stays consistent whether you’re looking at a low-res preview or full-res file.

CyberTimon:
Hi @d1anarene 

Thanks again for the update. 

The UX improvement with always showing the mask while dragging works good.

But I’m still noticing that the actual sharpening being applied doesn’t fully match what I’m seeing in the black/white mask preview. The mask preview looks correct, but the final sharpened result seems to affect different areas.

I think this might be related to how the threshold behaves across different resolutions. If the Canny/edge detection for the preview mask is being calculated on a downscaled preview image, that could explain the mismatch when the sharpening is later applied to the full-resolution RAW. Differences in scaling, interpolation, or gradient strength between resolutions might be shifting the effective threshold.

Maybe we need to either:
 - Normalize the threshold more strictly across resolutions (check how it's down with the blur shader itself) , or
 - Ensure the mask is generated from the same resolution that sharpening is ultimately applied to (or adjust the preview edge detection to better reflect the high res result (keep performance in mind)).

Could you take another look at how the preview mask and the final sharpening mask are derived relative to image resolution?

Thanks a lot and have a great evening,
Timon

GimleLarpes:
Is this ready for another review?

d1anarene:
Have not had much time to test it until now

R-Laine:
Can we get this moving again?

d1anarene:
right now some action from a third party crate fails. i will try to run it again later

CyberTimon:
Initial implementation done here (without the actual mask overlay): https://github.com/CyberTimon/RapidRAW/commit/344c55a194c2c756a5a03ecdc952d6d0f4063b7c

R-Laine:
will this be completed?

r3nor:
Also interested on this, hope it can get completed for a next release :)

PR #708, opened 2026-02-19
Added 2 clickable links in the issue pop-up

william-sy:
This config file adds 2 clickable links when you press "new issue" 
Making sure that users have a earlier encounter with the external libraries we ask them about in the bug template.


CyberTimon:
LGTM - thank you!

PR #710, opened 2026-02-19
feat(BottomBar): show selected image count relative to total images

savukadoo:
show selected image count relative to total images


CyberTimon:
Thank you! Looks good to me

PR #718, opened 2026-02-21
Persist main window size and position between launches

savukadoo:
This change adds the ability to persist windows size and position between RapidRAW app launches


CyberTimon:
Thanks a lot for the PR! 

I’ve polished this further by making sure the window remembers if it was maximized or in fullscreen. I also fixed the flickering issue during startup by waiting for the frontend to be ready before showing the window. Lastly, I tweaked the background colors so the app looks the same (in terms of brightness) whether transparency is on or off.

Kind regards,
Timon

PR #731, opened 2026-02-24
feat: Add DropDown search

GimleLarpes:
This PR adds simple search to the DropDown component. This is particularly useful for long dropdowns, such as the lens manufacturer dropdown. 
The input field is only made visible when the user starts typing, to prevent clutter in situations where search functionality is not needed (such as short lists, for example).

Example images:

image

image

Some things could likely be de-duplicated by using MainLibrary, however it's coupled to images to such a degree that doing a simple separate search implementation seemed like the better option.

The change in Input.tsx is motivated by the fact that the component uses a default `type="text"`, which was not indicated by its props, causing correct usage of the component to spawn errors.

Possible future improvements to DropDown: (I will probably tackle these next if this is merged)
- Replacing custom elements with premade UI components such as Button. 
- Implementing full keyboard navigation would also be good.

p.s.
This is my first contribution, but I am excited for the future of this project and would love to make more contributions! Is there a Matrix space or similar where the project is coordinated / things can be discussed? 
I'm mostly wondering if a more robust component system is a goal, since currently everything seems to have custom implementations for Buttons/Text/Spacing/etc. and I imagine consistent styling will become more difficult with time.

### DIFF
--- a/src/components/ui/Dropdown.tsx
+++ b/src/components/ui/Dropdown.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useRef, useMemo, useCallback } from 'react';
+import { useState, useEffect, useRef, useMemo } from 'react';
 import { AnimatePresence, motion } from 'framer-motion';
 import { Check, ChevronDown } from 'lucide-react';
 import Input from './Input';


CyberTimon:
Thanks a lot for this!! This is great work :) 

The DropDown search is a really nice addition, especially for longer lists like lens manufacturers, and I appreciate your idea to keeping the input hidden until it’s needed - perfect!!

You’re very welcome to join the Discord: https://discord.gg/cvFugZ2Hw8

It’s not super active right now, but it’s a place for discussion and coordination.

Will merge now & kind regards,
Timon

PR #733, opened 2026-02-24
Pull request templates

william-sy:
Add a pull request template to the repo

```
  _________________
< organic code YUM! >
  -----------------
         \   ^__^ 
          \  (oo)\_______
             (__)\       )\/\\
                 ||----w |
                 ||     ||
```

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,62 +1,43 @@
 ## Description
 <!-- Provide a clear and concise description of your changes -->
 
-
 ## Type of Change
-<!-- Mark the relevant option with an 'x' -->
-
-- [ ] Bug fix (non-breaking change which fixes an issue)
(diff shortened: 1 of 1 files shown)


GimleLarpes:
+1 for cowsay (Oh and the template is good too)

One question I do have is this: For my own workflow I often write the implementation, and then throw it into a few LLM's to see if they catch anything.
Would this fit into the "50%" category, or should a (for example) 10% category be added, that is effectively code spell-check?

william-sy:
@GimleLarpes, The wording suggest indeed its the 50% option, perhaps the percentages are not a good idea and it should just be check marks

YannBirba:
+1 Good idea

CyberTimon:
I rewrote some sentences and improved it here and there... Looks good to me now, will merge 💯

PR #734, opened 2026-02-24
Add default crop aspect ratio logic

savukadoo:
Added a settings drop down option to select a default aspect ratio. This default then gets applied to RAW files that dont have a aspect ration defined in the RRDATA file when using the crop tool. I also added in a few extra aspect ratio's that are very commonly used.


savukadoo:
Im closing this pull request as there are too many changes and I will rather not submit it

PR #740, opened 2026-02-25
fix show edited/original label

AntoniosBarotsis:
Does what the title says :)

image

Just started using rapidraw so not sure if the keybind used to be a `.` and this was left over from back then.


CyberTimon:
LGTM - it was a typo from older versions ;)

PR #738, opened 2026-02-24
refactor: Add Dropdown disabled state and clean up Dropdown usages

GimleLarpes:
This PR adds the disabled prop to `Dropdown`, as this prop makes sense and was used in some places (even though it didn't exist). This PR also  improves typing and cleans up some of the usages of Dropdown.

PR #745, opened 2026-02-27
refactor, fix: Slider improvements

GimleLarpes:
This PR fixes shift scroll not working, adds disabled state and refactors Slider to simplify the component. This PR also adds a generated docstring, I have checked that it is sufficient and technically correct.

This PR also updates usages of Slider.

Behaviour changes:
- Changes `onChange` return from `event[string | number]` to `number`
- Fix shift+scroll adjust
- ~Removed `handleRangeKeyDown`, as I do not see a reason for having it and caused buggy behaviour.~
- Hid default spinner buttons, custom ones should probably be added later.

~Depends on #738 (changes in Dropdown are from that pr, since I (partially) made it before branching)~


GimleLarpes:
This turned out to be a bigger diff than I had hoped for, but luckily all changes outside `Slider.tsx` are simple formatting or fixing typing.

CyberTimon:
With this PR checked out, the first time I opened the image I encountered the following error:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'toFixed')
    at Slider (Slider.tsx:262:91)
```

In addition:

* Shift + mouse wheel on the slider no longer updates the value.
* Pressing F (for fullscreen) while a slider is focused no longer toggles fullscreen; it simply refocuses the slider.

GimleLarpes:
> With this PR checked out, the first time I opened the image I encountered the following error:
> ```
> Uncaught TypeError: Cannot read properties of undefined (reading 'toFixed')
>     at Slider (Slider.tsx:262:91)
> ```
> 
> In addition:
> 
> * Shift + mouse wheel on the slider no longer updates the value.
> * Pressing F (for fullscreen) while a slider is focused no longer toggles fullscreen; it simply refocuses the slider.

Interesting, I'll look at this when I can. What browser/OS are you using? 
Regarding disabled global shortcuts, B for toggling visibility was very broken, so I disabled them. Though that might be a bug with just that shortcut. I'll look into reimplementing the functionality.

CyberTimon:
> Interesting, I'll look at this when I can. What browser/OS are you using?

Windows with Edge WebView 

> Regarding disabled global shortcuts, B for toggling visibility was very broken. Though that might be a bug with just that shortcut. I'll look into reimplementing the functionality.

Thanks! It isn't just related to B, it happens on all the shortcuts.

GimleLarpes:
~Also, where did you see the error and what were you doing to trigger it?  My guess is that there's some usage of the slider that sets value to undefined but type it as number. It would be amazing if you could find which of the slider usages throw the error! Since I can't reproduce for some reason.~

GimleLarpes:
Scrolling should now work for you, and I re-added the escaping of global keys. Something that should be fixed though, is that holding global keys down make the effect toggle every render. (That's a problem for later though!)

The Uncaught TypeError seems to be caused by something incorrectly passing undefined to slider as value. I'm having a hard time figuring out what though, as I'm unable to reproduce it and some parts of the project make liberal use of typing as any. Maybe you could provide the image that caused it, and say what modal/slider caused it?

**Edit:** I found it! The issue is in the MasksPanel. The CA was missing default values. I've fixed that now, but I also found an old bug lol

GimleLarpes:
I've now rebased this PR so it's now completely independent from #738 and they can be merged in any order.

GimleLarpes:
This should be ready for another review.

CyberTimon:
Thanks for the update!!

Just had a quick look, but moving a slider and then pressing F still doesn’t pass through to fullscreen - it just refocuses the slider :/
What do you think could cause this?

GimleLarpes:
> Thanks for the update!!
> 
> Just had a quick look, but moving a slider and then pressing F still doesn’t pass through to fullscreen - it just refocuses the slider :/ What do you think could cause this?

That is very weird as the code regarding that functionality is now identical to the code currently in `main`. Are you sure your local branch is up-to-date?

Here's what It looks like to me when I move the slider and press F:
[Screencast_20260301_221931.webm](https://github.com/user-attachments/assets/1b4417e0-8a9a-4878-8332-364e9ef5d685)

Other global keys also function like they did before in my testing.

GimleLarpes:
Could you re-check if you have the issue? I am still unable to replicate it by any of the following methods:
1. Clicking the slider and pressing a global key
2. Dragging the slider and then pressing a global key
3. Pressing a global key while dragging the slider
4. Tab-focusing the slider and pressing a global key

CyberTimon:
Main:

https://github.com/user-attachments/assets/eeba60ac-c376-45a7-81f0-db9b0799a7a9

Your PR:

https://github.com/user-attachments/assets/dcecd391-93f8-478b-bbc3-ee37c1aaa913

It's still not fixed. I will take a look at this myself if I find some time :)

Thanks!

GimleLarpes:
Hm, it might be a webview-specific issue. Would definitely be appreciated if you could take a look.

rkuykendall:
Hm, tried this on both Mac and Windows and was unable to reproduce the focus problem. Unsure if this PR is still aiming for merge, just trying to be useful and get my feet wet.

GimleLarpes:
Sorry for the late reply, It would be nice to get merged but one would probably have to rebase and basically redo the changes since IIRC the slider component has had significant changes. 

I might get to it when I have time, but I don't know when that'd be since I have a bit too much going on rn.

PR #755, opened 2026-02-28
chore: add lint/format tooling and CI

YannBirba:
## Summary
- Add npm scripts `lint`, `lint:fix`, `format`, `format:check` and ESLint/Prettier deps.
- Move to flat ESLint config (`eslint.config.js`) and add `.prettierignore`.
- Apply lint fixes and repo-wide formatting.
- Add CI workflow to run `format:check` and `lint`.
- Exclude `src-tauri/rawler` from lint/format checks.

## Tests
- `npm run format`
- `npm run lint:fix`
- (CI) `npm run format:check`
- (CI) `npm run lint`

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -50,7 +50,7 @@ module.exports = [
     rules: {
       'no-unused-vars': 'off',
       '@typescript-eslint/no-unused-vars': [
-        'error',
+        'warn',
         { argsIgnorePattern: '^_', varsIgnorePattern: '^_', caughtErrorsIgnorePattern: '^_' },
       ],
       // '@typescript-eslint/no-explicit-any': 'off',


GimleLarpes:
Adding tooling is great, and I myself want better linting, but I would say probably don't do a 21k line PR? IMO it would be better to just add the rules, and then over time files will conform to it.

YannBirba:
> Adding tooling is great, and I myself want better linting, but I would say probably don't do a 21k line PR? IMO it would be better to just add the rules, and then over time files will conform to it.

Yep, you're probably right, I will revert and just leave the eslint fixes for now !

GimleLarpes:
Also, do you think it would be a good idea to warn for unnecessary `any` type use? Looking at the current code base it seems like AIs love to do any type instead of robust typing, which isn't great. (My Slider PR already has problems due to this, as something somewhere passes undefined as any to it :( )

YannBirba:
> Also, do you think it would be a good idea to warn for unnecessary `any` type use? Looking at the current code base it seems like AIs love to do any type instead of robust typing, which isn't great. (My Slider PR already has problems due to this, as something somewhere passes undefined as any to it :( )

I agree with you. I think we have two options here:
- I can work on a strict ESLint/TypeScript configuration so we have a solid foundation for the IDE, but we would need to temporarily disable the CI checks while we fix the existing issues.
- Alternatively, we could keep the current setup and address the issues incrementally.

At the moment, ESLint reports 526 occurrences related to this problem.

For now, I’ve disabled the CI checks to avoid blocking progress. Let me know what you think would be the best approach moving forward.

CyberTimon:
Thanks for putting this together! This is a great addition to the project!!

I made one small change: I set no-unused-vars to warn instead of error for now. I don't want to block anyone while we all get used to the tooling. Once the codebase is mostly clean, we can promote it to error.

Merging this now. Appreciate the contribution!

PR #756, opened 2026-03-01
Improve preview performance and fix stale preview race condition

mwerneburg:
This PR addresses two performance issues and one stability/correctness bug found while profiling RapidRAW on a mid-range system  (i7-7700K, RX 6600).                                                                                                                     
                                                            
  1. Parallelize downscale_f32_image (performance)

  downscale_f32_image in image_processing.rs was implemented as a single-threaded nested loop using get_pixel() — a bounds-checked per-pixel accessor — with no parallelism despite rayon being available. For a 24MP image being downscaled to the preview resolution this involves tens of millions of accessor calls on a single thread.

  The implementation has been rewritten to use par_chunks_exact_mut over the output buffer (row-parallel) with direct flat slice access into the source image, eliminating both the per-pixel function call overhead and the single-threaded bottleneck. This is most noticeable during geometry operations (rotation, crop, lens correction) which invalidate the transform cache and trigger a fresh downscale.

  2. Fix stale preview race condition (correctness)

  preview-update-final events emitted by the backend carried no identifier indicating which image they belonged to. The preview worker thread processes jobs to completion before picking up the next one, so when navigating quickly between images a preview computed for the previous image could arrive after navigation and be displayed briefly for the new image — a visible "flicker back" to the previous photo.

The fix tags each preview-update-final event with the source image path. The frontend discards any event whose path does not match the currently selected image. Both emit sites (process_preview_job and generate_fullscreen_preview) have been updated.

Platform note: These issues were observed and the fixes tested on Linux (Ubuntu 24.04, Wayland, AMD RX 6600 with RADV Vulkan). The changes themselves are platform-agnostic — the Rust parallelisation and the event-tagging fix have no OS-specific dependencies — but they have not been tested on macOS or Windows.

### DIFF
--- a/src-tauri/src/image_processing.rs
+++ b/src-tauri/src/image_processing.rs
@@ -149,42 +149,45 @@ pub fn downscale_f32_image(image: &DynamicImage, nwidth: u32, nheight: u32) -> D
     }
 
     let img = image.to_rgb32f();
-    let mut out = Rgb32FImage::new(new_w, new_h);
+    let src: &[f32] = img.as_flat_samples().samples;
 
     let x_ratio = width as f32 / new_w as f32;
     let y_ratio = height as f32 / new_h as f32;
 
-    for y_out in 0..new_h {
-        for x_out in 0..new_w {
-            let x_start = (x_out as f32 * x_ratio).floor() as u32;
-            let y_start = (y_out as f32 * y_ratio).floor() as u32;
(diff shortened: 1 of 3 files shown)


CyberTimon:
Great work on this! I’m really glad to see that stale preview race condition finally sorted out!! I already tried to fix this so many times ;).. It makes a huge difference in the user experience. The Rayon optimizations are also fantastic. Thanks for the contribution!

PR #758, opened 2026-03-01
Fix panasonic g9ii lensfun alias

mwerneburg:
The Panasonic DC-G9 II reports itself as DC-G9 II in EXIF data (the English marketing name), but the lensfun database only contained the entry DC-G9M2 (the Japanese/internal designation used by Panasonic). With no alias bridging the two, camera body lookup fails entirely and lens correction profiles cannot be applied to images shot with this camera.

This change adds DC-G9 II as an English-language model alias for the existing DC-G9M2 entry, which already has the correct mount (Micro 4/3 System) and crop factor (2). No calibration data has been added or modified.

The fix was verified against a DC-G9 II shooting .rw2 files, rather than just inferred from the EXIF spec.

### DIFF
--- a/src-tauri/lensfun_db/mil-panasonic.xml
+++ b/src-tauri/lensfun_db/mil-panasonic.xml
@@ -137,6 +137,7 @@
     <camera>
         <maker>Panasonic</maker>
         <model>DC-G9M2</model>
+        <model lang="en">DC-G9 II</model>
         <mount>Micro 4/3 System</mount>
         <cropfactor>2</cropfactor>
     </camera>


bmx22c:
Seems like you have a commit that's unrelated to the PR: 20493bd633284f3a171944fbc5df84b73ca66bcf

mwerneburg:
Apologies. Cleaned up — the unrelated commit has been removed, the branch now only contains the lensfun alias change.

kmilos:
> The Panasonic DC-G9 II reports itself as DC-G9 II in EXIF data

No, [it does not](https://raw.pixls.us/). This "prettification" must be occuring somewhere else in RapidRaw (or even dnglab/rawler), and it still should use pristine "DC-G9M2" Exif data to _query_ the Lensfun database. (UI presentation of pretty names should be independent of this.)

Also note that none of the Panasonic models have these pretty aliases in the Lensfun database (see e.g. bottom camera list [here](https://lensfun.github.io/lenslist/)), as opposed to e.g. Canon, Pentax, or Sony... Why, I don't know... But if you do suggest something like this upstream, please do it consistently for all models.

PR #760, opened 2026-03-01
Improve export UX: remember format/folder, unify file naming

mwerneburg:
- Add lastExportPath and lastExportFormat to AppSettings so the chosen export folder and file format are restored across sessions
- Show the File Naming section for single-image exports (previously only visible in batch mode) and pre-fill the Save dialog filename using the filename template with {original_filename} substitution
- Both single and batch export paths now read and write lastExportPath, so the last-used directory is remembered regardless of export mode
- Library export panel likewise remembers the last export folder

### DIFF
--- a/src-tauri/src/image_processing.rs
+++ b/src-tauri/src/image_processing.rs
@@ -339,6 +339,100 @@ fn interpolate_pixel_with_tca(
     pixel_out[2] = sample_channel(bx, by, 2);
 }
 
+fn solve_generic_distortion_inv(r_target: f64, k_scaled: f64) -> f64 {
+    if k_scaled.abs() < 1e-9 {
+        return r_target;
+    }
+
+    let mut r = r_target;
+    for _ in 0..10 {
+        let r2 = r * r;
+        let val = k_scaled * r2 * r + r - r_target;
+        let slope = 3.0 * k_scaled * r2 + 1.0;
+        
+        if slope.abs() < 1e-9 { break; }
+        let delta = val / slope;
+        r -= delta;
+        if delta.abs() < 1e-6 { break; }
+    }
+    r
+}
+
+fn compute_lens_auto_crop_scale(params: &GeometryParams, width: f32, height: f32) -> f64 {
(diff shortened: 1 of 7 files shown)


CyberTimon:
Thanks for the PR @mwerneburg!

I've taken a look and found the following 2 things:

- Instead of saving the last export path/format, why not save a temporary snapshot of the last export preset? Or alternatively, persist the last export settings as a temporary preset? That way we capture the full state rather than just individual fields.

- Why did you remove the `{isBatchMode && (` check from ExportPanel? This causes duplicate "_Edited" name suffixes and other issues. We don't need two file naming fields in single export mode since the save file dialog already provides one.

Looking forward to your feedback & have a great start to the week,
Timon

mwerneburg:
Hi Timon,

Both points have been addressed in the updated branch:

1. Full export state snapshot
Rather than persisting just lastExportPath and lastExportFormat as individual AppSettings fields, the updated implementation saves a reserved ExportPreset with the id __last_used__. On mount, both ExportPanel and LibraryExportPanel read that preset and restore the full export state — format, quality, resize settings, watermark, filename template, and last output path — via the existing handleApplyPreset path. The preset is updated after each successful export. It is filtered out of the preset selector dropdown so users never see it.

2. isBatchMode guard restored
The {isBatchMode && ( check has been reinstated around the File Naming section in ExportPanel. You're right that it doesn't belong in single-image mode — the OS save dialog handles that. The template is still used internally to populate the suggested filename in the save dialog, but the section itself is no longer shown.

The ExportImportProperties.tsx conflict and the unrelated-commits issue have also been cleaned up via a rebase onto current main. I have been tripping over the various submissions, I will stop with multiple concurrent PR's.

CyberTimon:
Hi @mwerneburg 

Looks good to me, I’ll merge it now. The only last issue I’ve noticed is that `LibraryExportPanel` doesn’t rerender when `ExportPanel` changes the export settings - I’ll try to fix this on my own.

Thanks again for your contribution!

PR #762, opened 2026-03-01
Improve folder tree selection highlight visibility

mwerneburg:
The selected folder row used bg-surface as its background, which is a subtle elevation step barely distinguishable from the panel background. The hover state on unselected rows used bg-card-active, meaning hovering over a non-selected item was visually more prominent than the actual selection — making it very difficult to see which folder is active.

Change the selected row background to bg-accent/20 so it uses the theme accent colour at 20% opacity, giving a clear and consistent visual indicator that works across all themes.

### DIFF
--- a/src/components/panel/FolderTree.tsx
+++ b/src/components/panel/FolderTree.tsx
@@ -152,7 +152,7 @@ function TreeNode({
     <div className="text-sm">
       <div
         className={clsx('flex items-center gap-2 p-1.5 rounded-md transition-colors', {
-          'bg-surface': isSelected,
+          'bg-accent/20': isSelected,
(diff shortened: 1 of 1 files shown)


CyberTimon:
Thanks for the PR. 

I get your point but I think this is just in the nature of the dark theme:

image

Changing it at this single place would hurt the consistency of the other UI elements. What do you think?

mwerneburg:
I did consider it, but understood that it would be offset on all of the variants. Perhaps it was just the dark theme on Linux but I was having trouble seeing the highlight as it was in a moderately lit room. If it doesn't suit your design, by all means we can set it aside.

PR #763, opened 2026-03-01
Fix export hang on dialog

mwerneburg:
## Problem
                                                                                                                                           
The export status was set to `Exporting` before awaiting the file dialog. If the dialog hung (e.g. due to a Wayland window z-order issue) or failed silently, the UI was permanently stuck showing "Exporting (0/1)" with a spinner and no way to recover other than restarting the application.

## Fix

Move `setExportState(Exporting)` to after the dialog resolves and only when a valid path has been returned by the user. The UI now only enters the Exporting state once the export is genuinely underway. Applies to both `ExportPanel` and `LibraryExportPanel`.

Note this PR depends on the "export improvements" PR I submitted earlier.

### DIFF
--- a/src/components/panel/right/ExportPanel.tsx
+++ b/src/components/panel/right/ExportPanel.tsx
@@ -361,8 +361,6 @@ export default function ExportPanel({
       return;
     }
 
-    setExportState({ status: Status.Exporting, progress: { current: 0, total: numImages }, errorMessage: '' });
-
     let finalFilenameTemplate = filenameTemplate;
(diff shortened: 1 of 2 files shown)


CyberTimon:
Thanks for your PR. Will merge after you exclude the unrelated `Improve export UX: remember format/folder, unify file naming`-commit from this PR.

mwerneburg:
That should be sorted.

CyberTimon:
LGTM :) Thanks!

PR #766, opened 2026-03-01
feat: add CBM3D luma/chroma split denoising with mode toggle

mwerneburg:
Implements color BM3D (Dabov et al. 2007) by converting to YCbCr (ITU-R BT.601) before the BM3D pipeline, applying ~1.8x stronger sigma to the chrominance channels (Cb, Cr) and standard sigma to luminance (Y), then converting back. This targets the coarse color blobs typical of small-sensor chroma noise without further eroding luminance micro-detail.

A CBM3D/BM3D toggle button in the denoise modal lets users switch modes; the choice persists via AppSettings. Default strength lowered from 50 to 35, which empirically avoids muddying detail while still suppressing chroma noise.

Standard BM3D behavior (chroma_sigma_scale = 1.0) is fully preserved when CBM3D is off, so existing users see no regression.

You'll see a button on the denoise dialog. Click the small icon to enable CBM3D, I have tested this with several photos of outdoor and indoor settings and find it very suitable for dealing with noise in my micro-four-thirds photos while retaining sharp images. It is not considerably slower than the BM3D version.

### DIFF
--- a/src-tauri/src/denoising.rs
+++ b/src-tauri/src/denoising.rs
@@ -31,21 +31,16 @@ struct Bm3dParams {
 }
 
 impl Bm3dParams {
-    fn from_intensity(i: f32, use_cbm3d: bool) -> Self {
+    fn from_intensity(i: f32) -> Self {
         let val = i.clamp(0.001, 1.0);
-        let sigma = val * 80.0;
-        let lambda = 2.0 + (val * 2.5);
-        let dist = 3000.0 + (val * 20000.0);
-        // In CBM3D mode, chroma channels receive ~1.8x more smoothing than luma.
-        // This targets the coarser colour blobs typical of small-sensor chroma noise
-        // without further eroding luminance micro-detail.
-        let chroma_sigma_scale = if use_cbm3d { 1.8 } else { 1.0 };
-
         Self {
-            sigma,
-            hard_th_lambda: lambda,
-            max_dist_hard: dist,
-            chroma_sigma_scale,
+            sigma: val * 80.0,
+            hard_th_lambda: 2.0 + (val * 2.5),
+            max_dist_hard: 3000.0 + (val * 20000.0),
+            // Chroma channels (Cb, Cr) receive 1.8x stronger smoothing than luma (Y),
+            // suppressing the coarse colour blobs typical of small-sensor chroma noise
+            // without eroding luminance micro-detail.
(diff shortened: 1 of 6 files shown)


CyberTimon:
Thanks for the PR! A few notes from my testing:

* I don't see any difference & if it’s better, why not make it the default and remove the toggle?
* Could you share some side-by-side comparison examples?
* I’m seeing a green line on the bottom and right edges of the image after denoising...

mwerneburg:
I see the green line as well. I have a green background on my desktop, but that can't explain how I missed it. I'll look into that now. Also, I'm happy to redo without leaving the option for the prior version if you are. I can provide some examples, to me it's night and day (but then, I missed the green lines!). ;)

CyberTimon:
Thanks for the quick response!
I’d definitely like to see those side by side comparisons, and if you could provide a sample RAW file as well, that would be great so I can run some tests on my end.
Regarding the UI: if this new method is objectively better, let's just make it the standard behavior and remove the toggle entirely. There’s no need to include an option that produces worse results! ;)
Looking forward to the fix for those green lines.

mwerneburg:
Can do. I have a fix for the green lines en route. It looks like the switch to Cbm3d highlighted an existing issue where some pixels were being assigned a value of 0 which of course is invisible. Where shall I upload the raw and sample tif exports?

CyberTimon:
Thanks for the fix!
Just paste some 100% crops directly here in this issue. That’s sufficient for comparison :)

mwerneburg:
Hi! This morning (before an 11 hour work day) I found that RapidRAW as installed from the OS package on my Linux Mint system was producing 286MB files that couldn't be read by the OS photo browser. I submitted a fix for that this morning then spent a couple of hours looking at the results from both the BM3D version and the version I submitted with this PR. What I've come to realize is that even a fairly low 35% was far too much in a lot of cases, and that the resulting loss of detail was worse than the noise that was being treated. I am now submitting a version that takes a pass at the photo to pre-set the slider (it jumps a second or so after the dialog loads). It also re-applies fine texture to recover real features like hair and bushes etc. I will withdraw this PR.

CyberTimon:
Great! Can this PR be closed since https://github.com/CyberTimon/RapidRAW/pull/797 replaces it?

mwerneburg:
Yessir1

PR #783, opened 2026-03-02
Fix order of applying crop to distortion correction

christopher-kapic:
## Description
After further testing, when auto-cropping, images still looked warped. After investigation, I discovered that the order in which we applied the crop was incorrect. This PR corrects for that.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made

- Order of operations of distortion correction and auto-crop

## Testing

**Test Configuration:**
- **OS:** macOS Sequoia
- **Hardware:** Apple M1

## Checklist

## AI Disclaimer:

### DIFF
--- a/src-tauri/src/image_processing.rs
+++ b/src-tauri/src/image_processing.rs
@@ -667,17 +667,18 @@ pub fn unwarp_image_geometry(warped_image: &DynamicImage, params: GeometryParams
                     }
                 }
 
+                if auto_crop_scale > 1.0 {
+                    current_x = cx + (current_x - cx) * auto_crop_scale;
+                    current_y = cy + (current_y - cy) * auto_crop_scale;
+                }
+
                 let target_vec = forward_transform * NaVector3::new(current_x, current_y, 1.0);
 
                 if target_vec.z.abs() > 1e-6 {
                     let inv_z = 1.0 / target_vec.z;
-                    let mut src_x = target_vec.x * inv_z;
-                    let mut src_y = target_vec.y * inv_z;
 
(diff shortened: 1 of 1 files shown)


CyberTimon:
Could you please explain that in more detail? :) It looked correct in my tests..

christopher-kapic:
Sure thing!

First of all, here is an example of the current implementation:
image

And here is an example of my PR's implementation:
image


There is an important caveat: the lens with which I am testing is one that does not exist in the lensfun database (yet) and which has _heavy_ distortion at the corners (Sony 24-50 f/2.8). I tried following the documentation for creating a profile, but ran into several problems, so I vibe-coded a [CLI](https://github.com/christopher-kapic/lensfun-generator) which generates the lensfun profile. For distortion, it compares the raw file to the JPEG preview (since Sony applies corrections in-camera to the JPEG preview) and calculates the optimal c, b, and a values. I got _pretty decent_ results, but I don't think they are 100% perfect. I have some ideas on how to optimize this, but... I'm getting off-topic. The point is that my test images might not be the best for testing, but I don't think I have a lens that can produce a good test image, so if you have other heavily-distorted examples with which to test, it might be good to do more testing.

CyberTimon:
Thanks a lot for your explaination and for the sample comparison. I confirmed it on my end and saw the problem too (only seems to happen on wide angle lenses).

Will merge now! 🚀

PR #797, opened 2026-03-04
Feat smart denoising

mwerneburg:
## Type of Change
- [ x] New feature

## Changes Made
Extends BM3D denoiser with color-aware processing (CBM3D): the image is converted to YCbCr before denoising so that chrominance channels (Cb, Cr) receive 1.8× stronger smoothing than luminance, suppressing the coarse color blobs typical of small-sensor chroma noise without eroding luminance micro-detail. After denoising, a Gaussian blur of the original luma isolates fine texture (bark, hair, fabric), and a fraction of that residual is blended back into the denoised result, recovering detail that BM3D would otherwise smooth away.

Note that a new estimate_noise_level command analyses the image when the denoise dialog opens and pre-sets the intensity slider based on the measured noise sigma, so the starting point is tailored to each photo rather than a fixed default.

## Testing
- [x] I have tested these changes locally and confirmed that they work as expected without issues. Not quite DxO Photolab, but then what is. I have tested with underlit outdoor, twilight outdoor, and into-sun light to test the effect of the estimate_noise_level pass through the photo that sets the slider.

**Test Configuration:**
- Linux Mint
- i7-7700k and rx6600

## Checklist

## AI Disclaimer:

### DIFF
--- a/src-tauri/src/denoising.rs
+++ b/src-tauri/src/denoising.rs
@@ -806,46 +806,43 @@ fn gaussian_blur_1ch(data: &[f32], width: usize, height: usize, sigma: f32) -> V
         *k /= ksum;
     }
 
-    // Horizontal pass (each row is independent).
+    // Horizontal pass.
     let mut tmp = vec![0.0f32; width * height];
-    tmp.par_chunks_mut(width)
-        .enumerate()
-        .for_each(|(y, row_out)| {
-            let row_in = &data[y * width..(y + 1) * width];
-            for x in 0..width {
-                let mut val = 0.0f32;
-                let mut wsum = 0.0f32;
-                let x0 = x as isize - radius as isize;
-                for ki in 0..klen {
-                    let kx = x0 + ki as isize;
-                    if kx >= 0 && kx < width as isize {
-                        val += row_in[kx as usize] * kernel[ki];
-                        wsum += kernel[ki];
-                    }
+    for y in 0..height {
+        let row_in = &data[y * width..(y + 1) * width];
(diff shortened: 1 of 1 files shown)


CyberTimon:
Following up on https://github.com/CyberTimon/RapidRAW/pull/766, can you share some before / after 100% crops here?

CyberTimon:
I can now clearly see the improvements when using YCbCr. However, the estimator consistently returns a value of 10, and I’m not sure whether it’s actually necessary. It’s possible that YCbCr alone provides sufficient improvement.

We might consider setting the default to 10 or 15.

Additionally, I’m not observing any noticeable difference with the new Gaussian texture enhancement enabled or disabled.

mwerneburg:
I'll modify it to remove the estimation effort and set a low default. It's been an education in how much to use a denoiser!

SimonIT:
By the way, I just played with the EXIF tool with some RAWs and mine seem to have Noise Reduction Params in it: `Noise Reduction Params : 5 100 4 4 4 200 8 8 8 400 16 16 16 800 32 32 32 1600 64 64 64`. Maybe that's helpful?

CyberTimon:
Thanks for the update! Looks good to me.

mwerneburg:
> By the way, I just played with the EXIF tool with some RAWs and mine seem to have Noise Reduction Params in it: `Noise Reduction Params : 5 100 4 4 4 200 8 8 8 400 16 16 16 800 32 32 32 1600 64 64 64`. Maybe that's helpful?

This is a very interesting comment. We could easily use that to set the default slider value. Let me play with that and maybe make another PR.

PR #813, opened 2026-03-06
Fix crop handle clipping

mwerneburg:
## Description
The inner image canvas container had overflow-hidden applied alongside rounded-lg, which caused the browser to clip content at the rounded corners. This made ReactCrop's corner handles inaccessible when the crop selection reached the edges of the image. The outer editor container already applies overflow-hidden, so removing it from the inner container has no effect on pan/zoom clipping behavior.

## Type of Change
- [x ] Bug fix (depending on your point of view?)
- [x ] UI/UX improvement

## Changes Made
As above; couldn't see the corner of the image when trying a crop.

## Testing
- [x ] I have tested these changes locally and confirmed that they work as expected without issues

**Test Configuration:**
- **OS:** Linux Mint
- **Hardware:** i7-7700k and rx6600

## Checklist
- [x ] I haven't added unnecessary AI-generated code comments
- [x ] My changes generate no new warnings or errors

## AI Disclaimer:

### DIFF
--- a/src/components/panel/Editor.tsx
+++ b/src/components/panel/Editor.tsx
@@ -790,7 +790,7 @@ export default function Editor({
       </div>
 
       <div
-        className={clsx('flex-1 relative overflow-hidden', isFullScreen ? 'rounded-none' : 'rounded-lg')}
+        className={clsx('flex-1 relative', isFullScreen ? 'rounded-none' : 'rounded-lg')}
         onContextMenu={onContextMenu}
         ref={imageContainerRef}
       >


CyberTimon:
Thanks for the PR.

One thing I noticed while reviewing: this PR currently contains quite a few unrelated commits.

Have you tested these changes locally? I don't see any difference with overflow hidden removed or active.

CyberTimon:
Thanks for the PR.

One thing I noticed while reviewing: this PR currently contains quite a few unrelated commits.

Have you tested these changes locally? I don't see any difference with overflow hidden removed or active.

mwerneburg:
Dangit, I don't seem to be learning. I've rebased the branch — it should now show just the one commit.

On the visibility: you're right that it's subtle. The clipping only affects handles that sit right at the boundary of the inner container — so it's only apparent when the crop rectangle is dragged all the way to an edge of the image. With the crop inset from the edges, all four handles are fully inside the container and overflow-hidden has no effect. If you try expanding the crop to the full image and then attempt to grab a corner handle, the hit target should be noticeably easier to grab with this change. That said, if you're still not seeing a difference in that scenario, I'm happy to dig further — there may be a parent element doing the clipping instead, in which case this fix is in the wrong place. I have but one test bed.

P.S. Congrats on that latest review on Youtube, you're clearly getting some traction! ;)

PR #814, opened 2026-03-06
Fix batch adjustments

mwerneburg:
## Description
When multiple images were selected (Ctrl/Shift+click) and a slider was adjusted, the change was only persisted for the primary selectedImage. The other selected paths were never updated. I assume this was in error as the hooks to multiple image adjustments were present in the code.

The adjustment save path (useEffect watching adjustments) now calls apply_adjustments_to_paths for any selected paths beyond the primary image after each  non-dragging change. The backend merges adjustments key-by-key into each image's existing sidecar, so per-image settings that weren't touched in this session (e.g. individual white balance corrections) are preserved.

## Type of Change
- [x ] Bug fix (I suspect)
- [x ] UI/UX improvement

## Changes Made
As above.

## Testing
- [x ] I have tested these changes locally and confirmed that they work as expected without issues. It is glorious and undo works as well.

**Test Configuration:**
- **OS:** Linux Mint
- **Hardware:** i7-7700k and rx6600

## Checklist
- [x ] I haven't added unnecessary AI-generated code comments
- [x ] My changes generate no new warnings or errors

## AI Disclaimer:

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -282,6 +282,7 @@ function App() {
   });
   const [isSliderDragging, setIsSliderDragging] = useState(false);
   const dragIdleTimer = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const prevAdjustmentsRef = useRef<{ path: string; adjustments: Adjustments } | null>(null);
   const [isFullScreen, setIsFullScreen] = useState(false);
   const [isAnimatingTheme, setIsAnimatingTheme] = useState(false);
   const isInitialThemeMount = useRef(true);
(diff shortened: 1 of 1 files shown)


CyberTimon:
Thanks for the PR and for looking into batch adjustment.

One thing I noticed while reviewing: this PR currently contains quite a few unrelated commits (e.g., denoising changes and UI tweaks).

Regarding the adjustment handling: ideally the update should only modify the keys that actually changed instead of replacing the entire adjustments object. That helps avoid unintentionally overwriting existing per-image settings.

If you could split the unrelated commits and adjust that behavior, I can take another look.

CyberTimon:
Thanks so much for the update! This looks & works great.

Will merge now 🚀

PR #817, opened 2026-03-06
Add user-configurable font setting

preetpatel:
## Summary
- Adds a **Font** dropdown in General Settings with two options: **Poppins** (default) and **System Default**
- Poppins remains the default — this just gives users the option to switch
- Setting persists across sessions via `AppSettings`

## Motivation
Some users with accessibility needs or high-DPI/low-DPI displays may prefer their system font for better readability at small UI sizes. System fonts are also hinted and optimized per-platform (SF Pro on macOS, Segoe UI on Windows), so this can improve legibility for users who find that matters to them.

## Implementation
- Font is controlled via a `--font-family` CSS custom property referenced by Tailwind's `font-sans`
- Applied in the same `useEffect` that handles theme CSS variables
- Added `font_family: Option<String>` to the Rust `AppSettings` struct with `#[serde(default)]` for backwards compatibility

### DIFF
--- a/src-tauri/src/file_management.rs
+++ b/src-tauri/src/file_management.rs
@@ -304,6 +304,8 @@ pub struct AppSettings {
     pub sort_criteria: Option<SortCriteria>,
     pub filter_criteria: Option<FilterCriteria>,
     pub theme: Option<String>,
+    #[serde(default)]
+    pub font_family: Option<String>,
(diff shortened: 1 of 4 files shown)


CyberTimon:
Thanks a lot for your work! Looks good to me :)

PR #818, opened 2026-03-06
Bump All Minor Deps To Latest

RayJW:
## Description
New Rust version, npm complaining about vulnerabilities? This means it's time for another dependency clean-up. This upgrades all the minor stuff that doesn't affect functionality (or shouldn't). I will have more PRs coming trying to get some of the more tricky stuff in order.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made
- Update Rust MSRV to the latest
- Bump all Rust dependencies except `ort` (and thus `ndarray`)
- Bump all the minor npm dependencies

## Testing

**Test Configuration:**
- **OS:** CachyOS

## Checklist

## AI Disclaimer:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,10 +13,10 @@
         "@tauri-apps/plugin-os": "^2.3.2",
         "@tauri-apps/plugin-process": "^2.3.1",
         "@tauri-apps/plugin-shell": "^2.3.5",
-        "@uiw/color-convert": "^2.9.5",
-        "@uiw/react-color-wheel": "^2.9.5",
+        "@uiw/color-convert": "^2.9.6",
+        "@uiw/react-color-wheel": "^2.9.6",
         "clsx": "^2.1.1",
-        "framer-motion": "^12.35.0",
+        "framer-motion": "^12.35.2",
         "konva": "^10.2.0",
         "lodash.debounce": "^4.0.8",
         "lodash.throttle": "^4.1.1",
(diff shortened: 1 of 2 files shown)


RayJW:
Unless we want to explicitly specify the Rust version that should be used by the pipeline (which might actually be a good idea), this will have to rebuild once all the caches in the runners get Rust 1.94 I guess.

RayJW:
I think I found the reason for the build failure. RapidRAW pulls in `window-vibrancy` 0.7.1 while Tauri uses 0.6.0, while that itself is no problem, using 0.7.1 resolves `objc2` to version 0.6.4 which has a breaking change compared to Tauri, which resolves 0.6.3 for `objc2`. The way I see it we have a few options, I leave the decision to you @CyberTimon:

1. We remove `window-vibrancy` as a dependency since it's already part of Tauri, and we rely on [Tauris API](https://tauri.app/reference/javascript/api/namespacewindow/). From what I can gather that wouldn't change anything, but I don't have a Mac to test that on.
2. We pin `objc2` to 0.6.3 for the time being.
3. We downgrade `window-vibrancy` to 0.6.0 for the time being.
4. We patch the `Cargo.toml` to use `window-vibrancy` 0.7.1 for resolving Tauri.
5. We postpone any dependency upgrades until that is resolved.

I guess either 1 or 3 would be the cleanest options here, but I think all of them are viable in some way. I'd probably go with 1 since having less dependencies is usually preferable to avoid exactly such situations.

CyberTimon:
Hi @RayJW 

Thanks for digging into this and for the detailed analysis. I think option 1 makes the most sense here. Removing window-vibrancy and relying on Tauri’s API instead.

Functionality should be the same (if correctly added), right?

Thanks so much for all your help :)

Timon

RayJW:
> Functionality should be the same (if correctly added), right?

Yes, from my understanding. I'll open a separate PR for that, so it's a bit easier to review :)

RayJW:
I rebased this on #836 now, but it still won't build until #823 or another solution can be merged to make this always build with Rust 1.94, but I think it should be fine now.

CyberTimon:
Thanks! LGTM 💯

PR #838, opened 2026-03-10
Add ability to dock waveform into adjustments panel

savukadoo:
## Description
Added the ability to dock the waveform as it overlays on the image and gets in the way, so having it dock is a good way to have a clear space to make your edits as see the whole image and also view the waveform at the same time. I added it as a setting option, so that existing behavior remains and only if the users wants it docked then it will be a user defined setting.

## Type of Change
- [ x] New feature

## Changes Made
- Added a Option in the General Settings section to allow the use to toggle the waveform to dock in the adjustment section of the image editor
- If this toggle is off, the normal behavior is to overlay the waveform on the image canvas
- if this toggle is on, the behavior is to dock the waveform at the top of the Adjustments section, so its not overlayed on the image canvas 

## Screenshots/Videos

Screenshot 2026-03-10 at 22 17 08
Screenshot 2026-03-10 at 22 17 30

## Testing
- [ x] I have tested these changes locally and confirmed that they work as expected without issues

**Test Configuration:**
- **OS:** macOS Tahoe
- **Hardware:** Apple M2

## Checklist
- [ x] My code follows the project's code style
- [ x] I haven't added unnecessary AI-generated code comments
- [ x] My changes generate no new warnings or errors

## AI Disclaimer:
- [ x] This PR was handwritten with AI assistance (spell check, logic suggestions, error resolving)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -275,7 +275,7 @@ function App() {
   const [error, setError] = useState<string | null>(null);
   const [histogram, setHistogram] = useState<ChannelConfig | null>(null);
   const [waveform, setWaveform] = useState<WaveformData | null>(null);
-  const [isWaveformVisible, setIsWaveformVisible] = useState(false);
+  const [isWaveformVisible, setIsWaveformVisible] = useState(true);
   const [uiVisibility, setUiVisibility] = useState<UiVisibility>({
     folderTree: true,
     filmstrip: true,
@@ -1482,9 +1482,9 @@ function App() {
     }
   }, [uiVisibility, appSettings, handleSettingsChange]);
 
-  const handleToggleWaveform = useCallback(() => {
-    setIsWaveformVisible((prev: boolean) => !prev);
-  }, []);
+  //const handleToggleWaveform = useCallback(() => {
+//    setIsWaveformVisible((prev: boolean) => !prev);
+ // }, []);
 
   useEffect(() => {
     if (isInitialMount.current || !appSettings) {
@@ -4540,10 +4540,7 @@ function App() {
               isSliderDragging={isSliderDragging}
               isMaskControlHovered={isMaskControlHovered}
               isStraightenActive={isStraightenActive}
-              isWaveformDocked={appSettings?.dockWaveformInAdjustments ?? true}
-              isWaveformVisible={isWaveformVisible}
               onBackToLibrary={handleBackToLibrary}
-              onCloseWaveform={() => setIsWaveformVisible(false)}
               onContextMenu={handleEditorContextMenu}
               onGenerateAiMask={handleGenerateAiMask}
               onQuickErase={handleQuickErase}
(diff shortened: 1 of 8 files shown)


CyberTimon:
Hey @savukadoo, nice feature idea! 

A couple of things though:

Having this as a toggle buried in General Settings feels a bit out of the way... the user would have to leave what they're doing, open settings, flip it, and come back. It'd be much nicer to have a small dock/undock button directly inside the waveform display itself, similar to how the clipping indicator toggle works in the Curves panel. That way you can switch between docked and floating without leaving the editor.

Also for the setting to actually persist between sessions, you'll need to add the corresponding field to the settings struct in `file_management.rs`. Otherwise it'll reset every time the app restarts or the webview cache clears ;)

Thanks a lot for your work!
Timon

savukadoo:
Thanks for the feedback and agree with the buried aspect as its not clear. Just so I know , are you going to polish it off or do you want me to have a crack at that.

cheers

CyberTimon:
Yes, feel free to go ahead and implement that :)

Before you start though, do you think it might actually be better to keep the waveform docked all the time instead of having a toggle for it? The overlay tends to get in the way of the image a bit, so always docking it might make the editor cleaner and easier to work with.

savukadoo:
i would prefer it docked and placed all the time in the adjustments section, as it’s like the histogram idea i raised in the past , it’s a go to feature most photographers use all the time when editing and having it hidden by default. seems strange. i would like to in the future raise a PR to have a option in settings to choose your style, waveform or classic histogram, what are ideas and thoughts on that?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

savukadoo:
Hi Timon

I made all changes as requested and tested, however I am not sure if my branch was slightly out of date of some of the files as I see conflicts and Im not sure if they are related to me or new changes that have occurred on your side. I did my testing and everything looked good before submitting the changes.

There was alot of cleanup work as well just to remove references to the way the waveform was discplayed when selecting the icon method. I also removed the icon and heading / title from the waveform as its save space now, however you may want that back, however during my testing it just looked strange to have it there.

Screenshot 2026-03-14 at 10 22 46

CyberTimon:
Hey @savukadoo 

Sorry I actually ended up completely rewriting the waveform panel. It’s now more of an analytics panel that can display different scopes (e.g. vectorscope, etc.) instead of just the waveform.

Still thanks for your contribution which guided parts of this work. See commit https://github.com/CyberTimon/RapidRAW/commit/2fa909644e10578009bfab3b94244f64e1ece99a.

image

or

image

While doing this I also have complettely rewritten and optimized the backend calculation function.

Thanks!
Timon

savukadoo:
great work and glad I was some help. One ask, could you please add a more traditional scope tool - like the RGB histogram if you have the time as I do know that this is a common old school visual tool to help with tonal range. Most of us photographers have the same thing in-camera and use it to assist us with expsoure to some extent. Having a clipping detection is also very useful when using the basic adjustments. I see the histogram as a more vsual help and guide, where as the S-Curve tool is more for,  if you want to make more refined adjustments using the actual curve graph and its more interactive, so they have completely different user cases. 

I will be keen to test these new scopes out soon, let em know if there is anything else you want some assistance on and I can have a look.

PR #856, opened 2026-03-12
Fix: await settings save before relaunch on macOS (#835)

### LINKED ISSUE #835: BUG: Relaunch on MACOS not working
### What happened?

WHen changing a setting in the settings panel and the option to then save and relauncg the app is offered, I select it and the app never relaunches.

That the app would relaunch itself as the button stated when making changes that need the app to relaunch

1. Go into the setting section of the app
2. Choose an option that will offer a save and relaunch, like changing the processing backend
3. hit the save and relaunch

### Operating System

macOS Apple Silicon

### RapidRAW Version

1.5.1

### Additional information

I checked the code and the below might solve it, howver I cannot test on a DEV instance I am working on as I use 'npm run tauri dev'

`const handleSaveAndRelaunch = async () => {
  onSettingsChange({
    ...appSettings,
    ...processingSettings,
  });

  setTimeout(() => {
    relaunch();
  }, 200);
};
`

### Acknowledgement

jmadore:
Fixes #835

## Problem
On macOS, `handleSaveAndRelaunch` calls `onSettingsChange()` which internally fires `invoke(SaveSettings)` — an async write to disk. The promise was never awaited, so `relaunch()` was called before settings finished saving, causing the app to either fail to relaunch or restart with stale settings. The 200ms `setTimeout` was an unreliable workaround.

## Fix
- `handleSettingsChange` in `App.tsx` now returns the `invoke()` promise (`Promise<void>`)
- `onSettingsChange` interface updated from `void` → `Promise<void>` in `SettingsPanel.tsx` and `MainLibrary.tsx`
- `handleSaveAndRelaunch` now properly `await`s the settings save before calling `relaunch()`, replacing the fragile `setTimeout`

## Testing
Tested by changing the processing backend setting and clicking **Save & Relaunch** on macOS Apple Silicon.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1454,10 +1454,10 @@ function App() {
   );
 
   const handleSettingsChange = useCallback(
-    (newSettings: AppSettings) => {
+    (newSettings: AppSettings): Promise<void> => {
       if (!newSettings) {
         console.error('handleSettingsChange was called with null settings. Aborting save operation.');
-        return;
+        return Promise.resolve();
       }
       if (newSettings.theme && newSettings.theme !== theme) {
         setTheme(newSettings.theme);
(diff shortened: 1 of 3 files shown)


CyberTimon:
LGTM thanks a lot for the fix!

jmadore:
> LGTM thanks a lot for the fix!

I'm on macos so I am selfishly picking away at mac related issues. :-)

PR #855, opened 2026-03-12
Format All Rust Code

RayJW:
## Description
This once again formats everything Rust related to fix the workflow failing after the latest changes.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [X] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement
- [ ] Build/CI or Dependency update

## Changes Made
- Run `cargo fmt` over everything

## Testing
- [X] I have tested these changes locally and confirmed that they work as expected without issues

**Test Configuration:**
- **OS:** CachyOS
- **Hardware:** Ryzen 9, AMD graphics

## Checklist
- [X] My code follows the project's code style
- [X] I haven't added unnecessary AI-generated code comments
- [X] My changes generate no new warnings or errors

## AI Disclaimer:
Please state the involvement of AI in this PR:
- [ ] This PR is entirely AI-generated
- [ ] This PR is AI-generated but guided by a human
- [ ] This PR was handwritten with AI assistance (spell check, logic suggestions, error resolving)
- [X] This PR contains only blood, sweat, and coffee (AI-free)

PR #863, opened 2026-03-13
Replace rawler Submodule With Cargo Dependency

RayJW:
## Description
As discussed in #855, this replaces the `rawler` submodule with directly depending on it as a Cargo dependency. The way I see it, this has a few upsides:
- No `git submodule update --init` needed after cloning — Cargo handles it automatically (the same for updates)
- No `.gitmodules` file, no detached HEAD state confusion
- CI/contributors just run cargo build and it works

Possible cons, but they don't seem relevant here IMHO:
- You can't easily edit the dependency locally during development without `[patch.crates-io]` or a `path` override, but with the submodule already is treated as read-only, so no change there

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [X] Documentation update
- [ ] UI/UX improvement

## Changes Made
- Added `rawler` as a Cargo git dependency
- Removed the submodule
- Adjusted the workflows
- Changed the documentation accordingly 

## Testing

## Checklist

## AI Disclaimer:

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -46,4 +46,4 @@ jobs:
 
       - name: Clippy
         working-directory: src-tauri
-        run: cargo clippy -p RapidRAW --all-targets
+        run: cargo clippy --all-targets --all-features -- -D warnings


RayJW:
Quoting this here for the discussion:

> Hi @RayJW (/ @SimonIT)
> 
> Thanks for putting this together! I’m going to go ahead and merge this PR now as the code formatting is good and independent of the dependency discussion.
> 
> Regarding the suggestion to move rawler to a standard Cargo git dependency: While I agree that a git dependency is "cleaner" in theory, it would actually slow down my current development workflow. When I'm prototyping support for new image formats or specific cameras, I often need to make quick, iterative changes directly to the rawler files and build RapidRAW immediately to see the results.
> 
> Having to commit and push changes to a separate repository just to test a small tweak in the main app adds too much friction and overhead at this stage. What do you guys think? Let's discuss further in #863.
> 
> Thanks for the suggestion and the formatting work!
> 
> Timon

I think that's a totally valid reason and in the end it's your project :) There are some ways it could be handled, but I'll try to think about a better solution and opened #864 in the meantime to make the CI meaningful.

SimonIT:
I worked recently on my bachelor's thesis and was in the same situation with the different projects. Like RayJW suggested, I used `[patch.crates-io]` or `rawler = { path = "../../dnglab" }`. You can even use a different configuration file [config.toml](https://doc.rust-lang.org/cargo/reference/config.html) for the patches and not commit that file

RayJW:
> I worked recently on my bachelor's thesis and was in the same situation with the different projects. Like RayJW suggested, I used `[patch.crates-io]` or `rawler = { path = "../../dnglab" }`. You can even use a different configuration file [config.toml](https://doc.rust-lang.org/cargo/reference/config.html) for the patches and not commit that file

I think working with either of this would be best. Otherwise, Clippy will always complain so we'd have to decide on the lint failing for every CI run or never turning it on, both not very ideal. There is no way I could find to exclude just the rawler directory from the checks since it has to do the build in order to do all checks. Would one of these methods work for you @CyberTimon?

CyberTimon:
Hi everyone

Thanks for the idea! I think using a seperate patches toml file which I won't commit is the cleanest approach. 

Will merge it after I released v1.5.2 💯

CyberTimon:
Will merge after the rawler conflict is resolved. Thanks :)

RayJW:
> Will merge after the rawler conflict is resolved. Thanks :)

This is gtg now, only conflict was rawler being updated on main :)

CyberTimon:
Thanks! Will merge now 👍

PR #869, opened 2026-03-14
Added a edit icon to top right of edited image

savukadoo:
## Description
This PR adds a Pencil icon to edited images in the filmstrip and Library section of RapidRAW. This is helpful in identifying what images have been edited vs what images have not been images.

## Type of Change
- [ x] New feature

## Changes Made
- Added a Pencil icon to the Library section to any images that have been edited
- Added a Pencil icon to the Filmstrip section for any images that have been edited

## Screenshots/Videos
Screenshot 2026-03-14 at 22 07 16
Screenshot 2026-03-14 at 22 07 29

## Testing
- [ x] I have tested these changes locally and confirmed that they work as expected without issues

**Test Configuration:**
- **OS:** macOS Tahoe
- **Hardware:** Apple M2

## Checklist
- [ x] My code follows the project's code style
- [ x] I haven't added unnecessary AI-generated code comments
- [ x] My changes generate no new warnings or errors

## AI Disclaimer:
- [ x] This PR contains only blood, sweat, and coffee (AI-free)

### DIFF
--- a/src-tauri/src/file_management.rs
+++ b/src-tauri/src/file_management.rs
@@ -64,6 +64,144 @@ fn emit_thumbnail_cache_setup_error(app_handle: &AppHandle, path: &str, reason: 
     );
 }
 
+fn all_sections_visible_value() -> Value {
+    serde_json::json!({
+        "basic": true,
+        "color": true,
+        "curves": true,
+        "details": true,
+        "effects": true,
+    })
+}
+
+fn default_adjustments_value() -> Value {
+    serde_json::from_str(
+        r#"{
+            "aiPatches": [],
+            "blacks": 0,
+            "brightness": 0,
+            "centré": 0,
+            "clarity": 0,
+            "chromaticAberrationBlueYellow": 0,
+            "chromaticAberrationRedCyan": 0,
+            "colorCalibration": {
+                "blueHue": 0,
+                "blueSaturation": 0,
+                "greenHue": 0,
+                "greenSaturation": 0,
+                "redHue": 0,
+                "redSaturation": 0,
+                "shadowsTint": 0
+            },
+            "colorGrading": {
+                "balance": 0,
+                "blending": 50,
+                "highlights": { "hue": 0, "luminance": 0, "saturation": 0 },
+                "midtones": { "hue": 0, "luminance": 0, "saturation": 0 },
+                "shadows": { "hue": 0, "luminance": 0, "saturation": 0 }
+            },
+            "colorNoiseReduction": 0,
+            "contrast": 0,
+            "crop": null,
+            "curves": {
+                "blue": [{ "x": 0, "y": 0 }, { "x": 255, "y": 255 }],
+                "green": [{ "x": 0, "y": 0 }, { "x": 255, "y": 255 }],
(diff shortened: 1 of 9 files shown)


CyberTimon:
Hi @savukadoo 

I will take a look soon, but I have a quick question. 

Does it update correctly or does it become stale when you for example reset adjustments (it's in unedited state then). This was the reason https://github.com/CyberTimon/RapidRAW/pull/203 isn't merged yet.

Thanks!

savukadoo:
Il do some testing and confirm whether this occurs. Just to be clear so I understand the behavior we want.

- When a image is edited we want a icon to show that its been adjusted
- When a user selects the image and selects reset adjustments we want the edit icon to be removed
- What do we want to happen when a user rates and image or tags an image or color labels an image?
- Where do we want the edit icon to show, on the left top corner or the right top hand corner ?

Thanks

SimonIT:
Amd what if the user manually resets a slider? (For example drags it to 0 or double click and no other slider was changed)

CyberTimon:
The main issue here is that an 'edited' state is actually pretty tricky to define. A sidecar file gets created even if you just rate or tag an image, but we only want the icon to show up for actual visual changes. 

Also, if a user tweaks a slider and then moves it right back to the default value, the icon should ideally disappear. Keeping that state perfectly synced across the the editor and during batch actions (like pasting settings) is where most attempts at this end up feeling buggy or stale. We need to make sure the indicator is always telling the truth.

danschef:
Just an idea without knowing how the adjustments are actually stored in the sidecar files: 

What about storing a boolean value in the sidecar file that contains the "edited" state and which is refreshed every time the sidecar file is touched - lets say by simply computing the sum of all absolute edits (I guess they are stored somehow numerically). If it is zero, then set `edited=False` or even remove the sidecar file entirely.

SimonIT:
We could compare a static fresh rrdata struct with the actual one of the picture. If they are not equal, it is edited. This of course must be done on every change of any slider etc. This could happen in a function `is_default()`. I don't know the actual structure of the data. In case of multiple structs you could even create a trait for it.

CyberTimon:
On my phone rn, but an ideal place for this would be the save function. This way we don‘t have to compare on every slider tick.

savukadoo:
I think I have finally got to a point that has a potential working solution and I have done some testing on my side that looks like its the behavior we want , however I would need some more testing on your side to confirm that it is working as you expect it too.

Changes:

1. Moving a slider away from any of the default (0) value shows the pencil icon.
2. Returning the slider to default (0) value removes the pencil icon immediately.
3. Resetting adjustments removes the pencil icon.
4. Reopening an unedited image should keep the pencil icon hidden
5. Moved the pencil icon to the right and aligned it with the star rating if that is added.

[Screen Recording 2026-03-21 at 18.52.09.webm](https://github.com/user-attachments/assets/3c1f7e84-1173-4964-a084-9e4cacdb2609)

SebastianEggli:
maybe we could also add a filter to sort for edited photos.
This is something I've been missing.

CyberTimon:
All added here: https://github.com/CyberTimon/RapidRAW/commit/d278d0cfb8ba8fb841434be0dea7251ea58f5c75

Will appear in the next release :)

image

PR #899, opened 2026-03-20
TailwindCSS v4 Migration

RayJW:
## Description
Finally found time to revisit this one! Luckily the official migration tool already did most of the work. However, the way theming worked was completely broken, so I had to fix that and then some small stuff. PostCSS is not required any more as it is now all done with Vite.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [X] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made
- Ran the official migration tool
- Fixed some wrongfully changed code
- Migrated to the integration with Vite
- Fixed all broken theming

## Testing

**Test Configuration:**
- **OS:** CachyOS

## Checklist

## AI Disclaimer:
- [X] This PR was handwritten with AI assistance (spell check, logic suggestions, error resolving)

### DIFF
--- a/src/styles.css
+++ b/src/styles.css
@@ -1,4 +1,5 @@
 @import 'tailwindcss';
+@source not "../src-tauri";
 
 @theme {
   --color-bg-primary: var(--app-bg-primary);


CyberTimon:
Thanks for your work @RayJW!

The 2 things I've noticed:
 - Button's don't have a cursor style anymore. I found out this[ is actually intentional](https://dev.to/mmvergara/why-are-buttons-losing-the-pointer-cursor-the-shadcn-and-tailwind-debate-2j93), but still feels very weird to me. Can easily be fixed by adding this snipped to the `styles.css`:
 ```
   :where(button, [role='button']):not([disabled]) {
    cursor: pointer;
  }
 ```
  - Building takes a lot longer? I'm not sure why (it should be faster). On my Windows machine, I always run into `Frontend failed to report ready within timeout. Forcing window visibility.` and then have to watch a grey loading screen for like another 2-3 seconds. Start time with V3 was like 1-2 seconds and now it's 5-6+ seconds. Do you know what might cause this (just happens while developing, not on release builds)?
  
  Thanks!

RayJW:
> * Button's don't have a cursor style anymore. I found out this[ is actually intentional](https://dev.to/mmvergara/why-are-buttons-losing-the-pointer-cursor-the-shadcn-and-tailwind-debate-2j93), but still feels very weird to me.

Sure, I don't mind it personally, so I thought I'd stick with default behaviour, since the previous one was also default behaviour, but I've changed it back to be the same now :)

> * Building takes a lot longer? I'm not sure why (it should be faster). On my Windows machine, I always run into `Frontend failed to report ready within timeout. Forcing window visibility.` and then have to watch a grey loading screen for like another 2-3 seconds. Start time with V3 was like 1-2 seconds and now it's 5-6+ seconds. Do you know what might cause this (just happens while developing, not on release builds)?

That's weird. Are you building on Windows? For me on Linux build times are actually either faster or the same, but definitely not noticeably slower. I pushed a change to explicitly exclude the big directories not containing any relevant code from being walked on building. Apparently that can cause issues on Windows, where FS operations can be more inefficient, but disclaimer, that's a suggestion by AI and I can't currently confirm this since I don't have any Windows machine on hand.

CyberTimon:
Perfect! Looks good to me now. Thanks for the migration 🥇. The longer build times vanished after like 5-10 starts :)

PR #911, opened 2026-03-21
Mitigate large folder freeze during gallery startup

ByteMedic:
## Summary
This PR contains the branch work done to mitigate severe freezes that occurred when opening photo folders, especially with mixed RAW and large JPEG content.

## What changed
- reduced eager thumbnail work and batched frontend thumbnail updates
- requested thumbnails only for visible library and filmstrip ranges
- replaced base64 thumbnail IPC with cached asset paths
- reduced startup burst pressure during folder loading
- used embedded RAW previews where available
- skipped folder image counting when disabled
- deferred GPU initialization for gallery thumbnails
- refined startup thumbnail paths and added GPU diagnostics

## Notes
A branch summary was added in `README.bytemedic-fix-large-folder-freeze.md` to document the work commit by commit.


CyberTimon:
Hey @ByteMedic, thanks for putting this together!

I like several of the changes here and would love to keep them:

- Requesting thumbnails only for visible library and filmstrip ranges
- Reduced startup burst pressure during folder loading
- Using embedded RAW previews where available - great idea, but let's not hardcode which formats support it. Instead, just try to extract the embedded preview and if it fails instantly, fall back. That way we automatically support any format that happens to have one, without maintaining a list. Also make sure it's rotated corectly (doesn't work for portrait images rn)
- Skipping folder image counting when disabled
- Deferred GPU initialization for gallery thumbnails

Also, can we can remove the logic that checks whether images have non-visible adjustment changes. Looking through it, it adds a fair amount of complexity and feels prone to subtle maintenance bugs.

Overall, could we try to strike a better balance between the speedup gains and the complexity being introduced?

Thanks again for the effort here!

Timon

Issue #916, opened 2026-03-22
BUG: Transform Modal is completely broken in v1.5.2


CyberTimon:
### What happened?

Bug happens since I've implemented fine adjustment mode: https://github.com/CyberTimon/RapidRAW/commit/444790038b542cdfda1c61dc4403b93fdbe2d9bb

Seems related to some key up / down events not getting transmitted correctly in modals. Seems related to https://github.com/CyberTimon/RapidRAW/commit/421a75cd7f5e7ab03e432b88810f451ca2964d38

Sliders in lens correction modal also don't seem to work...

### What did you expect to happen?

:)

### Operating System

Windows 11

### RapidRAW Version

v1.5.2

### Acknowledgement

- [x] I have verified that my camera and lenses are supported by the external libraries listed above, or my issue is unrelated to camera/lens support.
- [x] I have searched all existing issues, both open and closed, and did not find a duplicate.

CyberTimon:
Distoration slider gives this: 
```
Preview transform failed invalid args `params` for command `preview_geometry_transform`: invalid type: null, expected f32
```

This is unrelated to the slider problem. Probably happens since https://github.com/CyberTimon/RapidRAW/commit/6e07119f27debf3dc8656b7ad80aebab70b0f457. @RayJW can you confirm?

RayJW:
> Distoration slider gives this:
> 
> ```
> Preview transform failed invalid args `params` for command `preview_geometry_transform`: invalid type: null, expected f32
> ```
> 
> This is unrelated to the slider problem. Probably happens since [6e07119](https://github.com/CyberTimon/RapidRAW/commit/6e07119f27debf3dc8656b7ad80aebab70b0f457). [@RayJW](https://github.com/RayJW) can you confirm?

I can confirm, it's the same on my end. I think the problem is actually, that they are missing the `step` prop with the new drag logic. Hence, the Rust error, because `snapToStep` then return NaN which obviously is not a valid type here. I've opened a PR that should fix the issue.

PR #919, opened 2026-03-22
Broken Sliders in Transform And Lens Correction

RayJW:
## Description
As discussed in #916, the sliders were missing a prop after the change to the way sliders were handled.

Fixes: #916 

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made
- Add `step` prop
- Fix the sliders in the transform modal being stuck to the cursor after dragging

## Testing

**Test Configuration:**
- **OS:** CachyOS

## Checklist

## AI Disclaimer:

### DIFF
--- a/src/components/ui/Slider.tsx
+++ b/src/components/ui/Slider.tsx
@@ -23,7 +23,7 @@ const Slider = ({
   min,
   onChange,
   onDragStateChange = () => {},
-  step,
+  step = 1,
   value,
   trackClassName,
 }: SliderProps) => {


CyberTimon:
Thanks! I'm on my phone right now but quick question: shouldn't we add a fallback for the slider so it doesn't return NaN if no step is defined, so it doesn't error out elsewhere if we miss the step prop?

RayJW:
> Thanks! I'm on my phone right now but quick question: shouldn't we add a fallback for the slider so it doesn't return NaN if no step is defined, so it doesn't error out elsewhere if we miss the step prop?

Yea, that seems like a good idea. I just set it to 1 as fallback, would you like me for the ones I've explicitly defined now to just use the fallback, or do you want to leave it explicit like this?

CyberTimon:
Looks good to me how it is set up now. Thanks!

PR #942, opened 2026-03-28
Make the brush same size so when zoomed in it covers a smaller area

Duecki1:
## Description
This PR makes the brush always have the same screen size instead of image size. This means when zooming in, the brush covers a smaller area on the image and when zooming out, it covers a bigger area. This practical because the user doesn't have to change the size all the time and can just zoom into the picture using the scroll wheel. (This is how the Lightroom brush works too)

## Type of Change
- [x] New feature

## Changes Made
- Brush size relative to the screen instead of the image

## Screenshots/Videos
https://github.com/user-attachments/assets/1bd407fe-986e-4e9e-b6af-6e6bc7a3be39

## Testing

## Checklist

## AI Disclaimer:

### DIFF
--- a/src/components/panel/Editor.tsx
+++ b/src/components/panel/Editor.tsx
@@ -846,6 +846,7 @@ export default function Editor({
               cursorStyle={cursorStyle}
               isMaxZoom={isMaxZoom}
               liveRotation={liveRotation}
+              zoomScale={transformState.scale}
             />
           </TransformComponent>
         </TransformWrapper>
--- a/src/components/panel/editor/ImageCanvas.tsx
+++ b/src/components/panel/editor/ImageCanvas.tsx
@@ -63,6 +63,7 @@ interface ImageCanvasProps {
   cursorStyle: string;
   isMaxZoom?: boolean;
   liveRotation?: number | null;
+  zoomScale: number;
 }
 
 interface MaskOverlay {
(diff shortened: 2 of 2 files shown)


CyberTimon:
@Duecki1 Perfect! Thanks a lot. This is a small but really helpful addition. If you are already working on the masking system, can you try to implement that it becomes negative / eraser while you hold down shift?

mmadesignerunknown:
Thanks Alot @Duecki1 
If possible, can we also add a functionality that if we hold Ctrl And use the Arrows, the brush size changes accordingly,
Like:
- Ctrl + Upwards Arrow Increases The Size,
- Ctrl + Downwards Arrow Decreases The Size,
 
It will be great if it can be implemented.
Thanks for the feature.😁

Duecki1:
> @Duecki1 Perfect! Thanks a lot. This is a small but really helpful addition. If you are already working on the masking system, can you try to implement that it becomes negative / eraser while you hold down shift? 

Yes, I will do that and the brush size shortcut.

mmadesignerunknown:
> Yes, I will do that and the brush size shortcut.

Hi, implemented both of those functions, it would be great if you could review those @Duecki1 .

The PR is open at https://github.com/CyberTimon/RapidRAW/pull/946

PR #961, opened 2026-04-03
Add EXIF columns to Library in List Mode (ISO, aperture, shutter, focal length) with sorting

savukadoo:
…and persistence

## Description
Adds dynamic EXIF columns to the Library list view with sorting and persistence across navigation.

## Type of Change
- [x] New feature
- [x] UI/UX improvement

## Changes Made
- Added selectable EXIF columns: ISO, Aperture, Shutter Speed, Focal Length
- Enabled sorting on all EXIF columns
- Persisted selected columns when switching between Library and Editor
- Fixed EXIF field mapping (Aperture ↔ Shutter)
- Standardized focal length display (mm)
- Implemented numeric shutter speed sorting (handles fractional values correctly)
- Fixed column/data alignment issues in list view

## Screenshots/Videos
Screenshot 2026-04-03 at 18 30 23
Screenshot 2026-04-03 at 18 30 36

## Testing

**Test Configuration:**
- **OS:** macOS Tahoe
- **Hardware:** Apple M2

## Checklist

## AI Disclaimer:

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -44,7 +44,7 @@ import {
 } from 'lucide-react';
 import TitleBar from './window/TitleBar';
 import CommunityPage from './components/panel/CommunityPage';
-import MainLibrary, { ColumnWidths } from './components/panel/MainLibrary';
+import MainLibrary, { ColumnWidths, LibraryColumnKey } from './components/panel/MainLibrary';
 import FolderTree from './components/panel/FolderTree';
 import Editor from './components/panel/Editor';
 import Controls from './components/panel/right/ControlsPanel';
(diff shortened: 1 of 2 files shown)


CyberTimon:
@savukadoo 
Thanks for your PR. In normal thumbnail view, I'm now also able to multi select EXIF sort keys. This seems buggy, what do you think?

savukadoo:
let me review that behaviour and see if it makes sense, i’ll have a look tonight

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

savukadoo:
I just had a play and I assume your question is today without this change you can use only one sort key and now with this change you can use multiple sort keys. I did some testing and you can remove and add and go back to one key, its sort of flexible in thumbnail view like it was before and now just has an extra dimension.

I have been thinking about PR's for RR a little and as you are probably going to see now and into the future a tsunami of them coming in for review as this app gains momentum and AI tech is evolving fast.  So, is there a better way that the community could vote? and I am only focusing this statement on major functionality that would effect workflows and major changes to the app. You ultimately or a small steering committee decide what goes up for vote, however that way the users can decide what they think is useful and beneficial.

I personally like this PR change, however its an opinion on my workflow and how it will benefit me, however like you , you may have a different perspective and that's a good thing as well.

CyberTimon:
Finally had time to [implement ](https://github.com/CyberTimon/RapidRAW/commit/e5e9fd3bb57611239557e2c5d91c06c1d83f5619)this, as we now cache EXIF data cleanly:

All:
image

On hover:
image

Off:

image

In lists:
image

Sort:
image

Thanks for the POC!

PR #964, opened 2026-04-03
feat: Add heic/heif support

### LINKED ISSUE #66: Add support for HEIC / HEIF image format
Hello,
I'm using iPhone 13. Default photo format is HEIC / HEIF format (which is based on h265 video format).
I'm using RapidRAW 1.2.3 on my Macbook pro M1 2020. It runs fantastic. But can't find any of my heic photos.
Support would be more than welcomed 🙂

SimonIT:
## Description

This adds support for editing HEIC and HEIF images. Support for the decoding hook isn't merged yet (imazen/heic#5). There is also limited colour depth at this stage, as only RGB8/RGBA8 is currently supported. But it's a start! 

Feel free to take a look at the decoding hook PR as well.

Closes #66

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made

- Include heic

## Testing

**Test Configuration:**
- **OS:** (e.g. Windows 11, macOS Sonoma, Ubuntu 24.04)
- **Hardware:** (e.g. Intel i7, Apple M2, Nvidia RTX 3060)

## Checklist

## AI Disclaimer:

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -17,6 +17,7 @@ dependencies = [
  "fuzzy-matcher",
  "glam 0.32.1",
  "half",
+ "heic",
  "hex",
  "image",
  "image-hdr",
@@ -198,40 +199,19 @@ checksum = "c3d036a3c4ab069c7b410a2ce876bd74808d2d0888a82667669f8e783a898bf1"
 
 [[package]]
 name = "archmage"
-version = "0.7.0"
+version = "0.9.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "61bd463aee29794a0318115ab857c2486842b28956c3f75365b304f1047b4758"
+checksum = "906d90739870dc217c19bac4b3d11541817679e875bc3ae03e66458f44ccac6b"
 dependencies = [
- "archmage-macros 0.7.0",
- "safe_unaligned_simd",
-]
-
-[[package]]
-name = "archmage"
-version = "0.9.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6fd400bfe079970d495bda0865afa8e3ba6f1604812f003c1ddcedc4d7a55a7e"
-dependencies = [
- "archmage-macros 0.9.5",
+ "archmage-macros",
  "safe_unaligned_simd",
 ]
 
 [[package]]
 name = "archmage-macros"
-version = "0.7.0"
+version = "0.9.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "02b01e2d131ffadba47bdff5d5e1d5d8a3b9743dc6b3d6ce4847ae2d6d306058"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 2.0.117",
-]
-
-[[package]]
-name = "archmage-macros"
-version = "0.9.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e73d1377bc9ebbed872f6a818decf8b360050183c641604078c0b5fc0f4a9189"
+checksum = "e2377968d27d15b6e6117b60026a12a76de7e79a40e1ac664158e2bc562d6d95"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -680,18 +660,15 @@ checksum = "5d20789868f4b01b2f2caec9f5c4e0213b41e3e5702a50157d699ae31ced2fcb"
 
 [[package]]
 name = "butteraugli"
-version = "0.7.2"
+version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "60d2d63e05316513a1385e0e3f680d24abb22d46c1eca4b20da9265f5d5d893c"
+checksum = "9c28a6704cdf610e0dfe761e2f289a3cac568db9128db12393f9463b4bf4fc76"
 dependencies = [
- "archmage 0.7.0",
+ "archmage",
  "imgref",
  "magetypes",
- "multiversed",
- "multiversion",
+ "rayon",
  "rgb",
- "simd_aligned",
- "wide 0.7.33",
 ]
 
 [[package]]
@@ -1529,9 +1506,9 @@ checksum = "66b7e2430c6dff6a955451e2cfc438f09cea1965a9d6f87f7e3b90decc014099"
 
 [[package]]
 name = "enough"
-version = "0.3.1"
+version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "65641ff0923e45cf34a4cb0115080cb7fc8bf4fd419c2207670d5722bab4c581"
+checksum = "521236efef7dcf4a95af2976e34b5b10780e8747eb78fa4742aa65b2ee189222"
 
 [[package]]
 name = "enumflags2"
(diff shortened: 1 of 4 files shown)


CyberTimon:
This is huge 😮. Last week I've checked and there was no pure rust heic / heif decoder. I wonder how this is possible, since there are many licensing issues regarding heic / heif. Can you verify if this is okay with their license of the heic / heif format in general? 

I will take a look at this in the next days. Also the decoder has 0 GitHub stars and no reputation, might be good to verify first.

Thanks a lot! I will also take a look at this after my holidays.

Timon

SimonIT:
Yes, it seems to be a brand new project — the first commit was on 22 January. Propably also the reason they have 0 stars. But they have managed to get it working:

RapidRAW 03 04 2026 21_30_20

Screenshot 03 04 2026 21_31_20

I'm no legal expert. It's licensed under AGPL, so it should be fine.

SimonIT:
https://github.com/nokiatech/heif/blob/master/LICENSE.TXT This seem to be the license for heif, but for HEIC I have no clue... But there seem to be other open source sdks, so there should be a way to use it

CyberTimon:
@SimonIT Thanks for putting this together! It's super cool to finally see a pure Rust HEIC decoder, and I really appreciate you taking the time to integrate it.

I did some research into the licensing, and we have a pretty big blocker regarding software patents. While the `imazen/heic` code itself is open source (AGPL), the actual math and algorithms used to decode HEVC (H.265) images are protected by thousands of patents owned by companies like Nokia, Apple, and Samsung. The Nokia license you linked earlier only covers the HEIF container format, not the actual HEVC video codec inside it.

This is a notorious trap in open source. It's the reason why browsers like Firefox still refuse to ship HEVC decoders. If I compile this into the RapidRAW binaries for Windows, macOS, or Flatpak, I'm technically distributing an unlicensed HEVC decoder, which is a massive legal risk.

I'd love to get the crate author's take on this. 

@lilith / @imazen Awesome work on the crate, a pure Rust HEVC decoder is a huge achievement. Since you offer it under AGPL and Commercial licenses, how do you handle the HEVC patent landscape (Access Advance, independent holders, etc.) for downstream projects? If a free open-source app like RapidRAW distributes compiled binaries that statically link your crate, does your license provide any patent coverage, or is the downstream distributor entirely responsible for navigating the patent pools?

@SimonIT What do you think about this? I'd love to hear your thoughts on the patent risk before we figure out the next steps for this PR.

Thanks!

SimonIT:
Maybe this document gives some insights? https://www.diva-portal.org/smash/get/diva2:1721030/FULLTEXT02.pdf AI seem to suggest that a warning in the README might be enough that users might need a license. But as I said, I'm no expert

CyberTimon:
@SimonIT 

If you read the abstract of the exact paper you just linked, it proves the exact opposite of what your AI suggested, and confirms the massive legal blocker. It explicitly states:

> *"Our analysis of legal and licensing conditions for use of the standard shows significant obstacles. **We find that it is impossible to obtain licences from patent holders that would allow for implementation and use of the standard in open source software.***"

In the eyes of patent law, putting a disclaimer or warning in the README does not protect the project. If we compile and distribute binaries (like Windows `.exe` files, macOS `.dmg`s, or Flatpaks) that contain an unlicensed HEVC decoder, we (the distributors) are committing patent infringement, regardless of what warnings we give to the end users. This is exactly why major open-source projects like Mozilla/Firefox refuse to ship HEVC decoders.

For now, we legally cannot merge and distribute this.

SimonIT:
That's really disappointing. We could hide it behind a feature flag, meaning users would have to compile it themselves to support HEIC/HEIF files. This way, we technically wouldn't be distributing the codec via exe, etc.

fstanis:
Hey @CyberTimon and @SimonIT, I looked into how some other apps do this, notably Chrome - it seems one patent-safe approach is to use a library to extract the HEVC stream from a HEIF file (which is not patent encumbered) and then delegate to either the OS (via [Media Foundation](https://learn.microsoft.com/en-us/windows/win32/medfound/h-265---hevc-video-decoder), [Video Toolbox](https://developer.apple.com/documentation/videotoolbox) or [Media Codec](https://developer.android.com/media/platform/transcoding)) or hardware directly (via VA-API, CUDA or Vulkan) for decoding for the actual HEVC -> bitmap decoding.

Obviously, this won't work in all cases since it requires either the OS or the hardware manufacturer to hold a HEVC license (some some of them [started disabling it](https://arstechnica.com/gadgets/2025/11/hp-and-dell-disable-hevc-support-built-into-their-laptops-cpus/) due to mounting costs apparently), but it's better than no support I guess.

Regarding the actual effort, luckily, ffmpeg actually supports all of this as of version 8 (including even HEIF containers). I tried prototyping just the HEIF-via-ffmpeg part and came up with [`ffheif`](https://github.com/fstanis/ffheif) and then also tried to just plug it into RapidRAW in #1158 and it worked (at least on Linux and macOS; I haven't tested Windows)! I created a PR if you'd like to try it out, note though that I have very little Rust experience so most of the Rust code is AI-generated - would appreciate someone more experienced to review.

CyberTimon:
@fstanis This is great. Does this need to have ffmpeg installed?

fstanis:
It doesn't - in the prototype I was able to link it statically (while disabling as many ffmpeg features as possible at compile time).

Ginushmal:
Not an expert on this in any way , but I have seen the open-source project Gimp supporting these file types, I think they use some way to get it done through the OS like @fstanis described.

alexdhill:
*I am not a lawyer*

From what I can gather the patents related to HEIC/HEIF are all on the HEVC/H.265 encoding/decoding. Looking at [imazen/heic](https://github.com/imazen/heic) they reference "optional native backends for Windows (Media Foundation), Apple (VideoToolbox), Android (MediaCodec), and Linux (VA-API)." This means if you build while **EXPLICITLY** not including the `rust-backend` decoder, you violate no patents.

Utilizing the native decoders means that
1. RapidRAW does not need to ship a HEVC decoder (ergo violating no patents)
2. All platforms currently supported *could* have a pre-existing, licensed decoder that will enable using HEIF/HEIC images - though there likely will be scenarios in which a Windows/Linux user doesn't have the right plugin/drivers installed

If someone wants to export to HEIC/HEIF however, it seems unsupported.

night199uk:
> 
> Utilizing the native decoders means that
> 
> 1. RapidRAW does not need to ship a HEVC decoder (ergo violating no patents)
> 2. All platforms currently supported _could_ have a pre-existing, licensed decoder that will enable using HEIF/HEIC images - though there likely will be scenarios in which a Windows/Linux user doesn't have the right plugin/drivers installed
> 
> If someone wants to export to HEIC/HEIF however, it seems unsupported.

Agree with this; utilizing rust heic and only including it's system backends by default (VAAPI, D3D, etc) would seem to bypass all the patent concerns as the user must have a supported/licensed heic backend on their system to get heif support. Users could enable the pure-rust backend if they compile for themselves. This seems a cleaner overall approach than the ffmpeg PR.

Decoding-only would seem beneficial at first given the popularity of the HEIC/HEIF format now, assuming the rust heic library will eventually get encoding via the same mechanisms it uses for decoding.

fstanis:
> This seems a cleaner overall approach than the ffmpeg PR.

I agree; FWIW, Rust's `heic` crate added system backends just days after I was playing around with the ffmpeg approach.

PR #969, opened 2026-04-04
New way of adding masks/components and a new floating mask hierarchy window.

Duecki1:
## Description
The masking panel was taking up so much vertical space on smaller devices with the big buttons to add masks. And on most screens there is more horizontal space, thats why I added the option for a floating mask hierarchy window too.

## Type of Change
- [x] New feature
- [x] UI/UX improvement

## Changes Made
- Added a floating mask hierarchy
- Removed the big buttons to add masks (Screenshots)

## Screenshots/Videos
Screenshot from 2026-04-04 22-53-34

Screenshot from 2026-04-04 22-53-27

## Testing

## Checklist

## AI Disclaimer:

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -331,6 +331,7 @@ function App() {
   const [activeAiPatchContainerId, setActiveAiPatchContainerId] = useState<string | null>(null);
   const [activeAiSubMaskId, setActiveAiSubMaskId] = useState<string | null>(null);
   const [maskHierarchyOverlayHost, setMaskHierarchyOverlayHost] = useState<HTMLDivElement | null>(null);
+  const [isMaskOverlayEnabled, setIsMaskOverlayEnabled] = useState(true);
   const [zoom, setZoom] = useState(1);
   const [displaySize, setDisplaySize] = useState<ImageDimensions>({ width: 0, height: 0 });
   const [previewSize, setPreviewSize] = useState<ImageDimensions>({ width: 0, height: 0 });
@@ -4958,6 +4959,7 @@ function App() {
               isLoading={isViewLoading}
               isSliderDragging={isSliderDragging}
               isMaskControlHovered={isMaskControlHovered}
+              isMaskOverlayEnabled={isMaskOverlayEnabled}
               isStraightenActive={isStraightenActive}
               onBackToLibrary={handleBackToLibrary}
               onContextMenu={handleEditorContextMenu}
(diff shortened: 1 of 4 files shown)


mmadesignerunknown:
Hi @Duecki1 ,

Thanks alot for this feature, while we are at Masks and it's UI, do you think that this feature should also be there:
- Double-clicking on an area of an image adjusts zoom to focus on that area.

What do you think about this. I just found that in Adjust tab, we have a feature that clicking on image focuses on that area, but since this single-click feature cannot be added to Masks (we have to create masks through click and drag), so i thought of implementing this as double-click. Same goes for Inpaint tab. It would be great if you can share your thoughts about it. :)
Thanks alot.

Regards,

CyberTimon:
Wow this is very cool! Thanks for implementing this @Duecki1. I wanted to ask if it's possible to keep the 2x3 grid the first time we open the Masks panel (when the mask controls / settings also aren't there yet). After the first mask creation, we never show them again.

Duecki1:
> Wow this is very cool! Thanks for implementing this @Duecki1. I wanted to ask if it's possible to keep the 2x3 grid the first time we open the Masks panel (when the mask controls / settings also aren't there yet). After the first mask creation, we never show them again.

Good idea, will do that 💯.

Duecki1:
@CyberTimon I've implemented the grid.

CyberTimon:
Thanks for the update, @Duecki1! 

The new layout is a great improvement. I’ve put together a small commit https://github.com/CyberTimon/RapidRAW/commit/07fab43c77474f521caca3455e93bfa60312cf98 that implements the core idea in a slightly simpler way. I felt that skipping the floating hierarchy would help keep the codebase lean while still capturing the space saving benefits.

Thanks for the idea and for the PR anyways ;)

Timon

Duecki1:
@CyberTimon 

wouldnt it be better to just display all of the components in one list instead of in two?
image


And also would there be any other way of using more horizontal space? Since right now a user has to scroll to the top everytime they want to change the selected mask. For me its quite annoying. 

Thanks tho for the implementation of some of the code :D.

Have a great evening, 
Linus

Duecki1:
and what about the reworked show overlay mechanism? That would be really cool too, since right now if a user tries to adjust a mask, they only see its red overlay and not the actual adjustments made to the picture. (And maybe the masking limit of 8, that's quite annoying too)

PR #1006, opened 2026-04-10
Fix UUID/Android Build

Duecki1:
This should fix the android build.

### DIFF
--- a/src-tauri/src/file_management.rs
+++ b/src-tauri/src/file_management.rs
@@ -28,7 +28,6 @@ use regex::Regex;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use tauri::{AppHandle, Emitter, Manager};
-#[cfg(any(target_os = "windows", target_os = "macos", target_os = "linux"))]
 use uuid::Uuid;
 use walkdir::WalkDir;
 


CyberTimon:
Thanks, missed that one ;)

Duecki1:
@CyberTimon May I ask, the AI Masks don't run the GPU right? Would it be possible to have a setting to change the model from a high performance option for devices with only a cpu to a high quality one for users with dedicated gpus?

CyberTimon:
@Duecki1 RapidRAW uses onnxruntime for inference. We need to get onnxruntime builds with GPU acceleration support to actually make use of the GPU. For Windows, we should use DirectML and for macOS, we can use Metal and on Linux, it's sadly not possible as far as I know. I've already tried to impelment this but got stuck in the dependency / library / architecture / compilation hell so I haven't tried any furter. Feel free to try implement this correctly :)

Duecki1:
> @Duecki1 RapidRAW uses onnxruntime for inference. We need to get onnxruntime builds with GPU acceleration support to actually make use of the GPU. For Windows, we should use DirectML and for macOS, we can use Metal and on Linux, it's sadly not possible as far as I know. I've already tried to impelment this but got stuck in the dependency / library / architecture / compilation hell so I haven't tried any furter. Feel free to try implement this correctly :)

Ohhh i see, thank you for the answer. I will take a look and check if there's any way to support it on linux without having to use many big dependencys :D

Have a great evening :)

Duecki1:
@CyberTimon another question if you don't mind. Do you know why the performance of RapidRaw is bad even on Ryzen 7 5700X3D with a 4070ti? For example the list view is really laggy with 400 pics, even tho the cpu and gpu are not being used to 100%. (Ubuntu on an SSD)

CyberTimon:
> @CyberTimon another question if you don't mind. Do you know why the performance of RapidRaw is bad even on Ryzen 7 5700X3D with a 4070ti? For example the list view is really laggy with 400 pics, even tho the cpu and gpu are not being used to 100%. (Ubuntu on an SSD)

Are you on a release build or are you running it with `npm start`? Also do you have Thumbnail Fit set to Fill Square or Original Ratio?
image

Duecki1:
> > @CyberTimon another question if you don't mind. Do you know why the performance of RapidRaw is bad even on Ryzen 7 5700X3D with a 4070ti? For example the list view is really laggy with 400 pics, even tho the cpu and gpu are not being used to 100%. (Ubuntu on an SSD)
> 
> Are you on a release build or are you running it with `npm start`? Also do you have Thumbnail Fit set to Fill Square or Original Ratio? image

ive build the newest version myself via "npx tauri build --debug", but the normal version does the same. Sometimes when I open the folder is just says "App unresponsive" and I need to click "Wait".

I'm using Original Ratio.

CyberTimon:
Try to set it to Fill Square as calculating the original ratio takes time and uses quite a bit more ressources. Also never do performance tests when running locally in the debugger, always test CI / release builds.

Duecki1:
> Try to set it to Fill Square as calculating the original ratio takes time and uses quite a bit more ressources. Also never do performance tests when running locally in the debugger, always test CI / release builds.

Will do and I've tested both the offical release and the debug version i have. Both had the issue. I think in the editor the performance gets bad when there are too many masks.

There are many weird behaviours.

For example I've noticed that when trying to add a brush mask to a subject mask, the performance of the brush mask is signifantly worse than on its own.

Brush Mask only: 

https://github.com/user-attachments/assets/95cd444e-3404-4e9f-9366-ece984ad360d

2 Subject Masks with Brush Mask:

https://github.com/user-attachments/assets/1b9d75de-d5d4-4d14-b5b9-063f543ad635

CyberTimon:
This is expected but can be fixed; the problem is that if we change **any** mask properties (the hashing algorithm is pretty basic atm), it transfers the full mask (which contains a huge base64 mask image) to the backend, every frame you drag the slider. We need to optimize this hashing so it actually just caches / hashes the actual base64 image and not just the whole mask object (which contains the huge image and the individual parameters which change often).

Duecki1:
@CyberTimon and another question if you don't mind, when would you want to publish the android port? I could later make a pr for the touch functionality. After that, the android port would be working for tablets. Only the UI for Phones would be needed.

CyberTimon:
> and another question if you don't mind, when would you want to publish the android port? I could later make a pr for the touch functionality. After that, the android port would be working for tablets. Only the UI for Phones would be needed.

@Duecki1 Yes, just a touch functionality PR would be excellent. This way we can quickly make RapidRAW work on tablets and we can maybe even get it into v1.5.4. I plan to release v1.5.4 by the end of the week. 

Thanks so much!

Duecki1:
> > and another question if you don't mind, when would you want to publish the android port? I could later make a pr for the touch functionality. After that, the android port would be working for tablets. Only the UI for Phones would be needed.
> 
> @Duecki1 Yes, just a touch functionality PR would be excellent. This way we can quickly make RapidRAW work on tablets and we can maybe even get it into v1.5.4. I plan to release v1.5.4 by the end of the week. 
> 
> Thanks so much!

And another thing, I don't really know how to make the ai masks work on android. Can I somewhere ask other people to help?

CyberTimon:
Can we continue this discussion on Discord. This merged PR isn't the right place :)

Duecki1:
> Can we continue this discussion on Discord. This merged PR isn't the right place :)

true, sorry :)

PR #1018, opened 2026-04-12
Add option to preserve export file timestamps from EXIF capture date

sevahul:
Hi!
Thank you for this project - I am very happy to conveniently post-process the photos in the open-source Ubuntu project.
I have made the change about the exported image timestamps for myself, and thought it might be useful also for others.

## Rationale
I wanted exported photos to be easily sortable by timestamp in external apps. In particular, this was an issue for me in Google Photos, where export ordering is much more reliable when file timestamps align with the original capture time.

## Summary
- add a new export option: **Set File Timestamps from EXIF Capture Date**
- wire the new preserveTimestamps setting through export presets and both export UIs
- when enabled, set output file timestamps (and exported mask files) based on EXIF capture datetime
- include backend support using the filetime crate

## Testing
- tested on Ubuntu 24.04
- verified that enabling the new option writes exported file timestamps from EXIF capture date
- verified that disabled behavior remains unchanged

-------------------

Main branch:
| Exporting | Resulting file |
| :---: | :---: |
| image | image |

Feature, enabled:
| Exporting | Resulting file |
| :---: | :---: |
| image | image |

Feature branch, the feature is disabled (matches the main branch):
| Exporting | Resulting file |
| :---: | :---: |
| image | image |

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -12,6 +12,7 @@ dependencies = [
  "bytemuck",
  "chrono",
  "fern",
+ "filetime",
  "futures",
  "futures-util",
  "fuzzy-matcher",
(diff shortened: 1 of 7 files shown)


CyberTimon:
Thanks @sevahul, this implementation looks a lot cleaner than before. Will merge now. Thanks for your contribution 💯 .

PR #1020, opened 2026-04-13
Remove 8 Mask Limit

Duecki1:
## Description
This PR removes the limit of only 8 Masks for each edit.

## Type of Change
- [x] Bug fix

## Changes Made
- Changed the Limit from 8 to 32

## Testing

## Checklist

## AI Disclaimer:

### DIFF
--- a/src-tauri/src/gpu_processing.rs
+++ b/src-tauri/src/gpu_processing.rs
@@ -5,7 +5,7 @@ use half::f16;
 use image::{DynamicImage, GenericImageView, ImageBuffer, Luma, Rgba};
 use wgpu::util::{DeviceExt, TextureDataOrder};
 
-use crate::image_processing::{AllAdjustments, GpuContext};
+use crate::image_processing::{AllAdjustments, GpuContext, MAX_MASKS};
 use crate::lut_processing::Lut;
 use crate::{AppState, GpuImageCache};
 
@@ -216,7 +216,6 @@ pub struct GpuProcessor {
     main_pipeline: wgpu::ComputePipeline,
     adjustments_buffer: wgpu::Buffer,
     dummy_blur_view: wgpu::TextureView,
-    dummy_mask_view: wgpu::TextureView,
     dummy_lut_view: wgpu::TextureView,
     dummy_lut_sampler: wgpu::Sampler,
     ping_pong_view: wgpu::TextureView,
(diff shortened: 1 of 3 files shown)


CyberTimon:
When merging this, we will make RapidRAW stop working for all these people & their devices:
image
Do you have a solution / fallback to this?

Duecki1:
> When merging this, we will make RapidRAW stop working for all these people & their devices: image Do you have a solution / fallback to this?

Texture Binding Arrays are not the same as a Texture 2D Array I think? Shouldn't this PR work on their devices? Or am I misunderstanding something? I don't have too much knowledge about this.

CyberTimon:
Hey @Duecki1

I did some more research and realized I was definitely a bit confused earlier! ;)

I was conflating the TEXTURE_BINDING_ARRAY feature (which caused the crashes on older hardware in the past) with a standard Texture 2D Array. Since your PR uses texture_2d_array<f32> and doesn't require the extra feature flag, it should be perfectly safe for all those devices I was worried about.

Your implementation is actually better for compatibility. I realized that my current hardcoded approach (the 16+ individual mask bindings) was putting us at about 23 total texture bindings. Since many older or mobile GPUs have a hardware limit of 16 bindings per shader, the current code was likely already breaking for some people. By grouping the masks into one array, you’ve brought us down to 8 bindings, which is much safer.

Great catch and thanks for the contribution. I'm merging this now!

Have a nice evening,
Timon

Duecki1:
> Hey @Duecki1
> 
> I did some more research and realized I was definitely a bit confused earlier! ;)
> 
> I was conflating the TEXTURE_BINDING_ARRAY feature (which caused the crashes on older hardware in the past) with a standard Texture 2D Array. Since your PR uses texture_2d_array<f32> and doesn't require the extra feature flag, it should be perfectly safe for all those devices I was worried about.
> 
> Your implementation is actually better for compatibility. I realized that my current hardcoded approach (the 16+ individual mask bindings) was putting us at about 23 total texture bindings. Since many older or mobile GPUs have a hardware limit of 16 bindings per shader, the current code was likely already breaking for some people. By grouping the masks into one array, you’ve brought us down to 8 bindings, which is much safer.
> 
> Great catch and thanks for the contribution. I'm merging this now!
> 
> Have a nice evening,
> Timon

I'm glad the code works :D. Thanks for merging it.

Thanks, you have a great evening too :).

PR #1027, opened 2026-04-15
docs: create contributing guide with architecture diagram

liamdasilva:
This PR adds a comprehensive CONTRIBUTING.md guide, including an interactive architecture diagram (Excalidraw) and updated project structure documentation. Documentation and images are moved to a dedicated /docs folder.


mmadesignerunknown:
Hi @liamdasilva ,
Thanks alot for this, although I was thinking that this whole guide should be moved in a separate project, just like:
- RapidRAW Docs: For user guides.
    - We already have this.
- RapidRAW AI Connector: For ComfyUI setup.
    - We already have this.
- RapidRAW Developer Docs: For developer guide.
    - This one; we don't have this.

Or move all three of them to this main repo, what do you think about this?

Regards,

liamdasilva:
> Hi @liamdasilva , Thanks alot for this, although I was thinking that this whole guide should be moved in a separate project, just like:
> 
> * RapidRAW Docs: For user guides.
>   
>   * We already have this.
> * RapidRAW AI Connector: For ComfyUI setup.
>   
>   * We already have this.
> * RapidRAW Developer Docs: For developer guide.
>   
>   * This one; we don't have this.
> 
> Or move all three of them to this main repo, what do you think about this?
> 
> Regards,

Hey there, happy to contribute! In general you would typically co-locate developer docs with the code itself since they are quite tightly coupled and are basically useless standalone without the code. For the other two guides, those are more user facing and and argument can be made for either case, I think its okay to leave them where they are for now and continue to evaluate.

One question - do those other docs typically need to be updated almost every time there is a code change? If so, then it might actually be better to co-locate them as it would indicate tighter coupling.

mmadesignerunknown:
Thanks, sure, as you wish :)
Yes I think the User Manual needs to be updated every time.
Although, it would be great of you could review the AI-Connector Docs and update them if needed.

Regards,

liamdasilva:
> Thanks, sure, as you wish :) Yes I think the User Manual needs to be updated every time. Although, it would be great of you could review the AI-Connector Docs and update them if needed.
> 
> Regards,

If the User Manual tends to need frequent updates, then I'd argue it should also live in this repo. Note that it would not bloat the size of the app itself since they would not be part of the build. As for the AI Connector docs, I don't actually see any, are you referring to [this](https://github.com/CyberTimon/RapidRAW-AI-Connector) repo? I think that repo is find to be standalone.

@CyberTimon I'll let you decide whether to move the user manual into this folder or not, we can have a dev guide sub folder and a user manual folder inside the parent docs folder. I can make the change if you're on board

mmadesignerunknown:
> As for the AI Connector docs, I don't actually see any, are you referring to [this](https://github.com/CyberTimon/RapidRAW-AI-Connector) repo? I think that repo is find to be standalone.

Yes, I am talking about this one, and I agree that it being standalone is fine, but don't you think that the complete software and it's related docs and tools should be in a one single repo rather being distributed # in multiple, it's just my opinion though and I will be glad to hear your thoughts about this.

Regards,

liamdasilva:
> don't you think that the complete software and it's related docs and tools should be in a one single repo rather being distributed # in multiple

yeah, an argument can be made for that, but moving the AI connector into this repo should probably be done in a separate PR if that's the route we want to go

mmadesignerunknown:
> > don't you think that the complete software and it's related docs and tools should be in a one single repo rather being distributed # in multiple
> 
> yeah, an argument can be made for that, but moving the AI connector into this repo should probably be done in a separate PR if that's the route we want to go

I agree. What do you think @CyberTimon?

PR #1029, opened 2026-04-15
reduce cpu and gpu usage

Fury7425:
updated the app to use less gpu and a bit less cpu while editing. (may feel a little bit laggy cause its not updating as soon as possible but probably enough although that is subjective)
upto -10% usage on my igpu when rapidly changing the values. and +3% to -8% usage on cpu.

Tested and didn't really see breaking (DIsclaimer: i dont really use keyboard shortcuts so idk if that broke)

## Test Configuration:
- **OS:** Windows 11
- **Hardware:** Intel core ultra 7 155h 16GB ram

## AI Disclaimer:
- This PR is AI-generated but guided by a human

| File | Change | Result |
|------|--------|--------|
| `src/hooks/useHistoryState.tsx` | Replaced `JSON.stringify` equality check with `shallowEqual()` (one-level key comparison) | Eliminates ~30KB serialization on every slider event at 60–120Hz |
| `src/components/ui/Slider.tsx` | `onChangeRef` dispatch batched through RAF via `pendingDispatchRef` + `dispatchRafRef`; flushed immediately on pointer up | `onChange` fires at most once per frame instead of on every `pointermove` event |
| `src/components/panel/editor/Waveform.tsx` | `HistogramView` path computation (`Math.max` + `map/join`) wrapped in `useMemo([red, green, blue])` | Path strings no longer recomputed on every parent render |
| `src/components/adjustments/Curves.tsx` | `getCurvePath(points)` result wrapped in `useMemo([points])`; component wrapped in `React.memo()` | Spline not recalculated unless control points change; component skips re-render on unrelated parent updates |
| `src/components/adjustments/Basic.tsx` | Wrapped in `React.memo()` | Skips re-render when unrelated sliders change |
| `src/components/adjustments/Color.tsx` | Wrapped in `React.memo()` | Skips re-render when unrelated sliders change |
| `src/components/adjustments/Details.tsx` | Wrapped in `React.memo()` | Skips re-render when unrelated sliders change |
| `src/components/adjustments/Effects.tsx` | Wrapped in `React.memo()` | Skips re-render when unrelated sliders change |
| `src/hooks/useKeyboardShortcuts.tsx` | 48-dep array replaced with `propsRef` pattern; dep array is now `[]` | Keydown listener registered once at mount, never torn down and re-added during slider drag |
| `src/hooks/useThumbnails.tsx` | Early `size === 0` guard at top of 150ms interval | `Array.from/filter` skipped entirely when visible queue is empty; interval self-cancels |
| `src/App.tsx` | `computeWaveform: !!isWaveformVisible && !dragging` | Backend skips waveform computation on every interactive drag frame; deferred until slider release |


CyberTimon:
Hi @Fury7425 

How exactly did you measure / profile the performance improvements? I don't really see a significant difference before / after and I'm a bit worried that this will just increase the code complexity without any noticeable benefit.

Thanks a lot for checking & for the contribution!

Timon

Fury7425:
I basically installed msi afterburner and rivatuner to get like the games fps and cpu gpu usage in real time but I just made it detect rapidraw also got the data and that's what I saw on my system.
But it probably is very small cause I am on a integrated gpu laptop. Probably on cpu max 6% and probably lower cause 155h is comparable to an i5 10gen(desktop) and I gpu is about a gtx 1080ti(also desktop) in dx12 so about max 8% and on a much modern system it probably only has a lower single didget difference. 
But for laptops a bit older it will probably help a little bit. 

This is a very small optimization in performance cause I really tried to not mess with the code much like all the changes except the keybinds are basically one or two letter differences. So it's really upto you to see if the little bit of a better experience on lower end and older hardware is worth it. I'd say why not but you are the maintainer and developer so yeah.

Fury7425:
Honestly I'd recommend accepting except the waveform and keyboards.

Idk why but now looking at it I forgot to fix the ais code to update the waveform in real time cause claude for some reason edited it so that it only updats after the change is made. and the keybind change did have large amounts of changes. But that's me and I made the pr so 🤷‍♂️

Fury7425:
BTW I forgot to save the results so here's a quick video recording to show you the results. https://quickshare.samsungcloud.com/75aPDCfeFNft

PR #1030, opened 2026-04-15
build: Turn all rust optimizations on

SimonIT:
## Description

This turns all rust optimizations on for release builds. This should make it faster, but the builds will take longer. The default was no optimizations

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made

- Set opt-level to 3

## Testing

**Test Configuration:**
- **OS:** Windows 11
- **Hardware:** i9 12900KF
## Checklist

## AI Disclaimer:

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -98,6 +98,7 @@ incremental = true
 opt-level = 3
 
 [profile.release]
+opt-level = 3
 codegen-units = 1
 lto = true
 strip = true


mmadesignerunknown:
Hi @SimonIT ,

I wanted to ask, that will all the builds take longer, like even the `"npm run tauri dev"`, or is it just for release builds?

Also, will it be an issue on devices with less than 16 GB RAM, (e.g. 8 GB)?

Thanks,

SimonIT:
No, it's in the release section (`[profile.release]`), so it's only for release builds.

I think you would have had a pretty bad time with 8GB of RAM also before this change. I just tried it with `cargo build --release` after a `cargo clean`: Before, it used 10 GB of RAM at peak for me, but after the change it's also around 10 GB. However, I only took those numbers by looking at the Task Manager, so it's not very scientific.

mmadesignerunknown:
> No, it's in the release section (`[profile.release]`), so it's only for release builds.
> 
> I think you would have had a pretty bad time with 8GB of RAM also before this change. I just tried it with `cargo build --release` after a `cargo clean`: Before, it used 10 GB of RAM at peak for me, but after the change it's also around 10 GB. However, I only took those numbers by looking at the Task Manager, so it's not very scientific.

Oh,thanks, 

Well, actually while using the `npm run tauri dev` command, it usually takes 5-10 minutes to launch the application if the `src-tauri` is unchanged. And if it is changed, it takes almost 15-20 minutes, or sometimes even 30. So yeah I mainly focus on frontend changes rather than backend.

CyberTimon:
Thanks for the PR @SimonIT! 

However, opt-level = 3 is actually the default setting for the [profile.release] profile in Cargo.
The 'no optimizations' (level 0) only applies to the dev profile (used for npm run tauri dev). Since this is in the release section, this change is redundant as Cargo is already using level 3 for release builds. You can verify this in the [Cargo Profiles documentation](https://doc.rust-lang.org/cargo/reference/profiles.html#release).

SimonIT:
Oh, I was wondering why I haven't felt any difference. I must have been looking at an older or different version of the documentation

PR #1034, opened 2026-04-15
Reverts to version 0.12 of reqwest to fix AI Masks on Android

Duecki1:
Reverts to version 0.12 of reqwest to fix AI Masks on Android


CyberTimon:
Did some minor changes so the changes just apply on android.

PR #1042, opened 2026-04-16
Rating assignment fix

R-Laine:
## Description
Fix inconsistent keyboard rating assignment by using the current image rating as the toggle baseline instead of the laggy adjustment snapshot. This keeps number-key shortcuts from unexpectedly clearing ratings when moving quickly between images.

When rapidly changing ratings via the num and arrow keys, issues such as the following would occur:
Rating removed (changed to 0) instead of updating to the new rating
Rating reverted soon after the press

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made
- Updated the rating handler to read the active image's rating from `imageRatings` before deciding whether a key press should toggle to zero.
- Kept the existing rating write path intact so the UI and backend still receive the same final rating updates.

## Screenshots/Videos
Not applicable.

## Testing

**Test Configuration:**
- **OS:** Windows
- **Hardware:** N/A

## Checklist

## AI Disclaimer:

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2735,12 +2735,8 @@ function App() {
         return;
       }
 
-      let currentRating = 0;
-      if (selectedImage && pathsToRate.includes(selectedImage.path)) {
-        currentRating = adjustments.rating;
-      } else if (libraryActivePath && pathsToRate.includes(libraryActivePath)) {
-        currentRating = libraryActiveAdjustments.rating;
-      }
+      const activePath = selectedImage?.path || libraryActivePath;
+      const currentRating = activePath ? imageRatings[activePath] || 0 : 0;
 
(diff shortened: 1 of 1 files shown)


CyberTimon:
Thanks for the PR! I really like the move to imageRatings for faster UI response, but the current logic has a side effect: it uses the active image's rating as the toggle baseline even when you're rating a different image via the context menu. This causes the rating to incorrectly toggle to 0 if the active image happens to share the same rating as the one you're clicking.

I've updated the logic to maintain the path validation while still using the imageRatings dictionary for that speed improvement.

Will merge now, thanks for your contribution :)

PR #1046, opened 2026-04-18
Revert "Rating assignment fix"

CyberTimon:
Reverts CyberTimon/RapidRAW#1042. Causes a lot of unexpected bugs when using multiple selected paths. Maybe @Thespacemanfil you can look into this? For you to reproduce: Try to select like 5 images and then press the key "4" to rate them with 4 stars. It will flash and flicker around endlessly...

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2757,17 +2757,13 @@ function App() {
         return;
       }
 
-      const activePath = selectedImage?.path || libraryActivePath;
       let currentRating = 0;
-      
       if (selectedImage && pathsToRate.includes(selectedImage.path)) {
-        currentRating = imageRatings[selectedImage.path] || 0;
+        currentRating = adjustments.rating;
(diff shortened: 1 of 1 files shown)


R-Laine:
Will take a look

R-Laine:
Found the issue(s):

currentRating was read from imageRatings and then overwritten by either adjustments.rating or libraryActiveAdjustments.rating, they're panel level values that can lag / differ from per-image rating state during multi-selection

Keyboard auto-repeat can re-fire rating updates quickly and trigger additional toggles while key is held causing the flickering

To solve them:

Use only imageRatings to get the currentRating in handleRate

Remove the dependency on adjustments.rating and libraryActiveAdjustments.rating in the callback

Ignore repeated keydown events in useKeyboardShortcuts.tsx for digit rating shortcuts, they shouldn't change at all until a different rating key is pressed

R-Laine:
https://github.com/CyberTimon/RapidRAW/pull/1053

PR #1048, opened 2026-04-18
fix(macos): init GPU context on main thread (#1047)

### LINKED ISSUE #1047: BUG: macOS panic when GPU initializes off main thread (get_metal_layer); crash recovery forces OpenGL
### What happened?

On macOS with Metal (or Auto, which uses Metal), the app can panic when the first GPU setup runs on a background thread. Thumbnail generation (`generate_thumbnails_progressive` uses a Rayon thread pool) and the preview worker both call `get_or_init_gpu_context`. That path creates a wgpu surface from the main window; Metal requires `get_metal_layer` on the main (UI) thread, so the process aborts with an error like:

`get_metal_layer` cannot be called in non-ui thread (via wgpu-hal / Rayon).

After that, the existing GPU crash recovery logic treats it like a driver failure and rewrites settings to use the OpenGL backend, which hides the real bug and can make previews unreliable on macOS.

Window/surface–backed GPU context for the main webview should be created on the main thread once. Preview and thumbnail work should only reuse that context from background threads, without panicking or triggering OpenGL fallback.

1. On macOS, set Processing → Processing backend to Auto or Metal (not OpenGL). Restart if the UI asks you to.
2. Open RapidRAW and open a folder so thumbnails start generating (or use the app in a way that triggers live previews early).
3. Observe a crash or abort; in logs, look for get_metal_layer cannot be called in non-ui thread (or similar wgpu/Metal threading panic).
4. Restart the app and check settings: Processing backend may now be OpenGL after GPU crash recovery.

### Operating System

macOS Apple Silicon

### RapidRAW Version

v1.5.3

### Additional info

- **OS:** macOS (Metal backend).
- **Scope:** This follows from **platform rules + where `get_or_init_gpu_context` is first called**, not from a broken local dev install; **release builds** can hit the same paths.
- **Relevant code:** `get_or_init_gpu_context` / `create_surface` in `src-tauri/src/gpu_processing.rs`; callers include the **preview worker** (`process_preview_job`) and **thumbnail** paths in `src-tauri/src/file_management.rs` (e.g. `spawn_blocking` / Rayon). Crash fallback: `GPU Driver crash detected` / `processing_backend` → `gl` in `src-tauri/src/lib.rs` setup.
- **Possible fix:** After `window_builder.build()` in `setup`, call `get_or_init_gpu_context` **once on the main thread** so background threads only see a cached context. Optionally narrow crash recovery so a **threading panic** does not force OpenGL on macOS.

### Acknowledgement

Flohhhhh:
Hi after setting up my dev environment for the first time, I was unable to get passed this crash/issue.

I believe that this commit earlier today may have been the cause: [“implement direct wgpu renderer”](https://github.com/CyberTimon/RapidRAW/commit/5edd9fb17fa446ae3399dfc682aa782a6dfa4a11) or similar changes?

## Summary
Pre-initializes `get_or_init_gpu_context` on the **main thread** immediately after the main window is built in `setup`.

On macOS, Metal requires `get_metal_layer` / surface creation on the UI thread. Thumbnail generation (Rayon) and the preview worker call `get_or_init_gpu_context` from **background threads**; if they run first, the process can panic and crash recovery incorrectly forces OpenGL.

## Changes
- After `window_builder.build()`, call `get_or_init_gpu_context` once (non-Android) so later callers only clone the cached context.

Fixes #1047

Assisted by [Cursor](https://cursor.com)

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -4891,6 +4891,17 @@ pub fn run() {
 
             #[cfg(not(target_os = "android"))]
             {
+                // Metal requires creating the window surface on the main thread. Preview and
+                // thumbnail work runs on background threads; initialize once here so they only
+                // clone the cached GpuContext (avoids panic + erroneous OpenGL fallback).
+                let app_state = app.state::<AppState>();
+                if let Err(error) = get_or_init_gpu_context(&app_state, app.handle()) {
+                    log::warn!(
+                        "GPU pre-initialization failed (editing and thumbnails may be degraded): {}",
+                        error
+                    );
+                }
+
                 if let Ok(config_dir) = app.path().app_config_dir() {
                     let path = config_dir.join("window_state.json");
                     if let Ok(contents) = std::fs::read_to_string(&path) {


CyberTimon:
Thanks a lot for your fix! LGTM! 

PS: Haven't had time to test the wgpu renderer on macOS but I expected some bugs. Glad you could already solve that one :)

Issue #321, opened 2025-09-15
enhancements: Library page - multiple improvements


avoiceofreason:
These all relate to the Library / thumbnail page

1.When you change Thumbnail Fit from "Fill Square" to "Original Ratio" you get a blurred edging to fill the square. It hurts my eyes. Prefer just a black background. 

2.If more images are copied into the folder you are using the icons are not updated by default. Be good to have a "Refresh Icons" button up top to rescan folder and populate for missing images.

3.At the bottom are two buttons: a:Reset All Adjustments and b:Export Selected Images. Please change the button icon for "Reset All Adjustments" to something that looks like a "delete" icon. I pressed it in haste thinking it was a refresh icons button and inadvertently deleted the edits to several files. Also add a "Are you Sure Y/N" option box as this is a pretty destructive button.

4.Add at the bottom a "Backup all Sidecars files" button so that all RR sidecars files are zipped into a file in that folder so edits can be backed up from time to time. Use a file sequencing approach so multiple backup files are created over time.

william-sy:
A few thoughts:

3.1. A delete Icon might not be right either, As that might give the 'I will delete the file' feeling. Currently it is a reset icon. Perhaps the reset icon with a clock in it might be a alternative to give you "you are going back in time" vibes. 
3.2. Another thing could be changing the text color to red or orange to indicate a destructive action and perhaps a horizontal separation between the 2 actions might also be a solution. 
3.2. Not sure on a confirmation prompt as you already had to right click. So see 3.2 perhaps a better visual separation might be a compromise

4.1. These files add up in space though 10kb or less for fresh images and 1MB or less for edited pictures. There are users here that have a 2000+ image gallery they want to edit. assuming a 50/50 split that is 1GB worth of files. Copying a large amount of small files is not great for performance so you need a logic to only do this for files with changes. (which complicates things)

CyberTimon:
> When you change Thumbnail Fit from "Fill Square" to "Original Ratio" you get a blurred edging to fill the square. It hurts my eyes. Prefer just a black background.

I'm sorry that you don't like it, but I've heard many reports that people actually like it a lot.

> If more images are copied into the folder you are using the icons are not updated by default. Be good to have a "Refresh Icons" button up top to rescan folder and populate for missing images.

I'm not sure if this is really needed or if it just clutters the UI. You can just press on the folder again in the sidebar to rescan the folder.

avoiceofreason:
[...]

PR #1049, opened 2026-04-18
ui: Adjust some icons for better clarity

Flohhhhh:
Hi this is my first *opinion* based suggestion! I am very interested in both UX and UI flows as a frontend engineer and I found the icons on the tab buttons to be a little bit vague/misleading. Hopefully these changes make sense! I understand if you prefer the old ones.

Updates the right panel switcher to use more specific Lucide icons per tab:

- **Crop** — `Crop` (was `Scaling`)
- **Masks** — `Blend` (was `Layers`)
- **Inpaint** — `Paintbrush` (was `BrushCleaning`)
- **Presets** — `SwatchBook` (was `Bookmark`)
- **Export** — `FileInput` (was `Save`)

- Also adjusted the misleading "Reset Adjustments" icon based on comments in #321 to be "Eraser" instead of "RotateCCW" which gave the impression of a refresh button

Info and Adjust tabs unchanged.
image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,6 +19,7 @@ import {
   CopyPlus,
   Edit,
   FileEdit,
+  FileInput,
   Folder,
   FolderInput,
   FolderPlus,
@@ -28,7 +29,6 @@ import {
   RefreshCw,
   RotateCcw,
   Star,
-  Save,
   SquaresUnite,
   Palette,
   Tag,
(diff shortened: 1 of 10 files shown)


CyberTimon:
Great! Looks good to me, except the Masks one. I much prefer the layers icon as it symbolizes multiple layers of adjustments over the base image (e.g. masks). The blend icon looks more like mixing 2 images into each other.

What do you think? @Flohhhhh

Flohhhhh:
@CyberTimon Fine with me, not very attached to that one. Personally I think there's nothing in Lucide right now that feels amazing for masks.

(I thought it looked like radial masks)

CyberTimon:
Thanks a lot! Looks good to me now 💯

PR #1053, opened 2026-04-19
Fixing the ratings fix

R-Laine:
Reverts CyberTimon/RapidRAW#1046

fixes my pull request

issue(s):

currentRating was read from imageRatings and then overwritten by either adjustments.rating or libraryActiveAdjustments.rating, they're panel level values that can lag / differ from per-image rating state during multi-selection

Keyboard auto-repeat can re-fire rating updates quickly and trigger additional toggles while key is held causing the flickering

To solve them:

Use only imageRatings to get the currentRating in handleRate

Remove the dependency on adjustments.rating and libraryActiveAdjustments.rating in the callback

Ignore repeated keydown events in useKeyboardShortcuts.tsx for digit rating shortcuts, they shouldn't change at all until a different rating key is pressed

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -4946,8 +4946,8 @@ pub fn run() {
                 let window_for_handler = window.clone();
                 let pending_state_for_handler = pending_window_state.clone();
 
-                window.on_window_event(move |event| match event {
-                    tauri::WindowEvent::Resized(_) | tauri::WindowEvent::Moved(_) => {
+                window.on_window_event(move |event| {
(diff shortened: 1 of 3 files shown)


R-Laine:
Can't figure out the lint issue, if anyone has any knowledge on it

CyberTimon:
@Thespacemanfil 

Thanks for the update. The correct way to actually fix this isn't to build guards around it; instead should move the ratings out of the adjustments struct and correctly integrate it into the image's metadata. I implemented this real quick and it's working correctly now, without the need for any guards or similar hacks: https://github.com/CyberTimon/RapidRAW/commit/6fc0e9c8309e22785d36871befa07635180efcb8

Thanks for understanding :)

PR #1055, opened 2026-04-19
ui: Add alert dialog before resetting adjustments

Flohhhhh:
## Description
I added a confirmation modal/alert dialog step for resetting adjustments since some users mentioned in #321 that they were accidentally performing this destructive action and losing their work

## Type of Change
- [x] UI/UX improvement

## Changes Made
Added a reusable `promptResetAdjustments()` function to request a confirmation before performing the action.

## Screenshots/Videos
image

## Testing

**Test Configuration:**
- **OS:** (e.g. Windows 11, macOS Sonoma, Ubuntu 24.04)
- **Hardware:** (e.g. Intel i7, Apple M2, Nvidia RTX 3060)

## Checklist

## AI Disclaimer:

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4238,8 +4238,8 @@ function App() {
         confirmText: 'Reset',
         isOpen: true,
         message: isSingle
-          ? 'This removes all edits for this image and restores defaults. Rating is kept.'
-          : `This removes all edits for ${count} images and restores defaults. Ratings are kept.`,
+          ? 'This will remove all edits from this image and reset the image to its original state and cannot be undone.'
+          : `This will remove all edits from ${count} images and reset the images to their original states and cannot be undone.`,
         onConfirm: () => handleResetAdjustments(pathsToReset),
         title: isSingle ? 'Reset adjustments?' : `Reset adjustments on ${count} images?`,
       });
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4226,6 +4226,27 @@ function App() {
     ],
   );
 
+  const promptResetAdjustments = useCallback(
+    (explicitPaths?: Array<string>) => {
+      const pathsToReset = explicitPaths ?? multiSelectedPaths;
+      if (pathsToReset.length === 0) {
+        return;
+      }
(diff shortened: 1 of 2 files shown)


R-Laine:
Wouldn't the undo button be sufficient? Don't see the need for the kind of alert after the first 3 clicks or so

Flohhhhh:
I would say it's generally a better goal to prevent mistakes rather than relying on an "oops I didn't understand" button.

Undo is currently scoped to per-file edit history, which is a clear and predictable model. Supporting undo for this action would require expanding that scope to “undo any action in the app,” which I think would only hurt UX in the long run. This would also be the first non-adjustment undo action available in the app.

`(It could end up like Lightroom where simply changing files with the arrow key is an undo step 😭 )`

Also I assume to provide an undo we would rely on a toast like "✔️ 5 images were reset" and an undo button on it? This is very fragile, users may not understand what they did fully before the window expires.

I don't think there's any concern about this slowing users down too much, this is a very low frequency high impact destructive action which is the perfect use case for an alert dialog. I expect only 30-50% of users would ever encounter this dialog, and probably only half of those would actually click "Reset". 

(I'm not sure what you meant by "3 clicks", I'm mostly worried about the one click reset button present on the bottom of the library page)

An optional improvement would be to store "Reset Adjustments" as a history item in a file's undo history so theoretically if they bulk reset some images, then visited that image in edit view they could use undo to step back and undo the reset. (I created an issue regarding this since I noticed undo state is not persistent right now #1059)

CyberTimon:
Moved confirm option into a submenu:

image

I think this is the best tradeoff to prevent accidental clicks but keep it fast & clean.

Flohhhhh:
@CyberTimon this doesn't help with this button on the library view though right?
image

CyberTimon:
@Flohhhhh I removed that button from the bottombar. This button was actually just a button in the initial days used for quick testing ;)

Flohhhhh:
@CyberTimon That works! 

I will admit I'm not the biggest fan of the confirm action being in a submenu in general, but it's a suitable compromise.

Thanks :D

PR #1062, opened 2026-04-21
Adds the gen folder and hopefully fixes the signing

Duecki1:
Adds the gen folder and hopefully fixes the signing


Duecki1:
The APK works and has the RapidRaw Icon

CyberTimon:
LGTM - Thanks for the fix!

PR #1074, opened 2026-04-23
Fix white balance & straight line tool

Duecki1:


CyberTimon:
LGTM, thanks for the quick fix!

Issue #1082, opened 2026-04-24
BUG: Mouse scroll not zooming


ysaric:
### What happened?

When I bring a photo up to edit, mouse scrolling does not zoom the photo in and out. It will do a very large-jump zoom if I hold the Ctrl key down while using the mouse scroll wheel, but mousing over the picture it gives the "+" mouse pointer where left clicking does a modest zoom which then turns the mouse cursor to "-" and clicking again zooms back out.  Because it is the mouse cursor exhibiting this behavior it is quite tricky to try and get a screen capture of what that looks like. My mouse scroll wheel tries to scroll the photo up/down when the photo is zoomed in larger than the screen, like in a Word document. When the photo is entirely in the editing frame the mouse scroll wheel has not effect.

### What did you expect to happen?

Expected the mouse scroll wheel to zoom in/out as its function.

### How can we reproduce this?

Open a photo in editing
Mouse over the picture
Try to use the scroll wheel to zoom in/out

### Operating System

Windows 11

### RapidRAW Version

1.5.4

### Additional information

If you have trouble reproducing the issue I could do a screen recording that shows the behavior on my system.

### Acknowledgement

- [x] I have verified that my camera and lenses are supported by the external libraries listed above, or my issue is unrelated to camera/lens support.
- [x] I have searched all existing issues, both open and closed, and did not find a duplicate.

CyberTimon:
I'm aware of this bug. Some OS translate mouse guestures differently and the current detection isn't releiable enough. 

The better solution to the auto detection seems a simple settings section, where you can configure how you want the Canvas to behave:

<img width="691" height="299" alt="Image" src="https://github.com/user-attachments/assets/78b71110-14b1-456a-920a-9adec445dd7d" />

Will appear in the next release, see https://github.com/CyberTimon/RapidRAW/commit/c9d2700001833794858e6ca3ea42d303db85487a.

PR #1078, opened 2026-04-23
Numpad controls (SP-3000 style)

R-Laine:
## Description
Implemented numpad hotkeys for rapid image adjustments, inspired by traditional film scanner workflows. Users can now adjust RGB/CMY color channels, contrast, and density using the numpad keys, with configurable step sizes and two adjustment modes.

This was requested in https://github.com/CyberTimon/RapidRAW/issues/96

## Type of Change
- [x] New feature

## Changes Made
- Added settings interface for numpad configuration with mode (digital/film), enter key behavior, and configurable step sizes
- Extended adjustments interface with color offsets for film-style color grading
- Updated keyboard shortcuts hook to handle all numpad keys with:
  - Digital mode: Temperature/Tint adjustments
  - Film mode: Full RGB/CMY color offsets with density-rich shadows
  - Contrast adjustments
  - Reset function
  - Enter key modes: Next, Instant Export, Skip & Move
- Created settings panel section with:
  - Enable/disable toggle
  - Mode switch (Digital/Film)
  - Enter key behavior selector
  - Adjustable step sizes for exposure, contrast, and RGB/CMY
  - Key mapping reference table
- Integrated with main app to pass numpad settings and adjustments to the keyboard hook

## Screenshots/Videos
image
image
image
image

## Critical Issue: Film Mode Not Integrated with Backend

The Film mode (RGB/CMY color offsets) is currently **not functional** because it lacks backend integration.

### Problem
- Frontend stores `filmColorOffsets` (cyanRed, magentaGreen, yellowBlue) in the adjustments state
- Numpad shortcuts successfully modify these values in the frontend
- However, the Rust backend ([GlobalAdjustments](cci:2://file:///c:/Users/fil/Documents/GitHub/RapidRAW/src-tauri/src/image_processing.rs:1191:0-1273:1) struct in [image_processing.rs](cci:7://file:///c:/Users/fil/Documents/GitHub/RapidRAW/src-tauri/src/image_processing.rs:0:0-0:0)) and WGSL shader do not include these fields
- No shader logic exists to apply these color offsets to the rendered image

### Impact
- Film mode adjustments are saved but **do not affect image rendering**
- Digital mode works fully (uses existing temperature/tint which are already in the backend)
- Users will see no visual change when using Film mode numpad keys

### Resolution Required
To make Film mode functional, backend changes are needed:
[...]

PR #1083, opened 2026-04-24
Feature: Configurable Keybindings for RapidRaw

sefalkner:
## Configurable Keybindings for RapidRaw

This PR introduces configurable keybindings/controls for RapidRaw. As a requirement for that, the `handleKeyDown` was refactored going from many nested if/else statements to a simpler to edit matching strategy.

This should give users the control they want, and we developers don't break their workflows as easily.

**Motivation:**

- Make it your software - People might have different ideas of a good workflow
- Make migration from other software  easier
- Internationalization - Record physical key positions instead of symbols
- Custom Periphery - Maybe for the future
- Safety - E.g., allow users to unassign delete
- The big trackpad rugpull - Issues like #1082 already came up for the changed mouse settings, your new panel for customization fits very well into the controls page

This should maybe get priority over #1078, as the numpad can be bound to tasks now.

**For Users:**

UI-wise, not a lot has changed (with full intent). To the user, they can now simply click on a keyboard on the shortcuts (now controls) page and record a keybinding.

Users can also "unassign" a function as described above. If keybindings already exist, they are marked as conflicting but not auto-removed. If conflicting keys are kept, the first one in the `handleKeyDown` of the conflicting ones fires, should be pretty safe.

**For Developers:**

I tried my best to not bloat the code unnecessarily while allowing for easy extension later.

Key Recording records physical key position through `event.code`, more robust on different language keyboards than the actual key.

There are two types of keybindings, builtin and user configurable. Builtins are always checked first, meaning the user can't override them.

New keybinds just get added to `KEYBIND_DEFINITIONS`. Each one has a condition (`shouldFire`) and a handler (`execute`). On key press, builtin guards run first; if none match, the pressed combo is looked up in the user-configured map and its handler runs.

**Needs Feedback/Discussion/Testing:**

- **macOS**  
  Since I don't have access, I'd appreciate some testing there, especially the ctrl+delete shortcuts.
- **numlock**  
  Numlock is currently not respected, as we record the physical key positions.
- **Locale Handling**  
  Right now, keybinds are displayed on a QWERTY keyboard. This is a compromise many packages do, even vscode, but maybe we can do better in the future. Sadly this is not fully supported yet: https://developer.mozilla.org/en-US/docs/Web/API/Keyboard_API
- **Primary/secondary**  
  For now, there is just a single keybind per action, but giving a secondary should be easy, do you prefer that?
- **Smarter Conflict detection**  
  For keybind that only fire in certain situations, the conflict detection should not fire if they never could collide with another specific keybind.
- **Brush Size changes**  
  I was not able to get it running, the regression might actually be not in the keybind code
- **Where should the keybind definition live**
  Right now it's in src/utils/keyboardUtils.ts, which is a bit underwhelming given its importance

Let me know your feedback and ideas, I'd be fine we keep this open until all points are discussed.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Screenshots/Videos

Screenshot From 2026-04-24 19-20-28

## Testing

**Test Configuration:**
- **OS:** Ubuntu 24.04
- **Hardware:** AMD AI HX 370

## Checklist

## AI Disclaimer:
- [x] This PR is AI-generated but guided by a human, in no world I'd reformat these keybinds like a caveman ;)


CyberTimon:
This looks absolutely amazing!! Really clean refactor and a huge usability win. The configurable keybindings and the move away from nested conditionals make the system much easier to extend and reason about.

Looks good to me, will merge now! 💯

PR #1097, opened 2026-04-26
[Android] Fix FS operations, enable all supported formats import, enable library backup, fix LUT

waxm134:
## Description
- moved library path to /sdcard/Android/media
- Android library import can now handle all supported RAW formats
- Android LUT file picker fixed

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made
- Android library now lives in  /sdcard/Android/media/io.github.CyberTimon.RapidRAW/.library. This allows backup of the library with non-root file managers
- Import file picker on Android does not use file extension filters as Android expects MIMI types, but not all supported RAW file types have a consistent Android MIME type, therefore the frontend checks the file type of the selected files and displays an error toast if the file is unsupported
- LUT file picker uses the same approach as import file picker on Android. While fixing this, I also discovered that we only supported lower-case LUT extensions until now, fixed that as well
- LUT files get cached on Android in  /sdcard/Android/media/io.github.CyberTimon.RapidRAW/.lut_cache. This is necessary, because with the current minimal permission set the app is only allowed to access the LUT file when the user actually selects it from the file picker. With the caching any later attempt to read the file again will be done from the app-owned .lut_cache folder which does not have the strict file access permission issues

## Testing

**Test Configuration:**
- **OS:**  Ubuntu 24.04, Android 16
- **Hardware:** Intel i5, Xiaomi Pad 7 

## Checklist

## Additional Notes
- we must use hidden folders in favor of .nomedia files because hidden folders don't trigger the android media scanner. With normal folders and .nomedia files the media scanner locks access to any .jpeg/png etc. files it finds in the folders even if they are not shown in the gallery. 
- The code implements a few workarounds to make sure that we don't need higher file access permissions on Android. This is a bit messy in some places. It should be discussed whether we want to request the external storage access permission in future so that we don't have to cache all files coming from outside the app

## AI Disclaimer:

### DIFF
--- a/src-tauri/src/file_management.rs
+++ b/src-tauri/src/file_management.rs
@@ -587,8 +587,7 @@ pub fn parse_virtual_path(virtual_path: &str) -> (PathBuf, PathBuf) {
     (source_path, sidecar_path)
 }
 
-#[cfg(target_os = "android")]
-fn is_android_content_uri(path: &str) -> bool {
+pub fn is_android_content_uri(path: &str) -> bool {
     path.starts_with("content://")
 }
 
@@ -670,7 +669,7 @@ fn parse_android_uri<'local>(
 }
 
 #[cfg(target_os = "android")]
-fn resolve_android_content_uri_name(uri_str: &str) -> Result<String, String> {
+pub fn resolve_android_content_uri_name(uri_str: &str) -> Result<String, String> {
     let vm = unsafe { JavaVM::from_raw(android_context().vm().cast()) }
         .map_err(|e| format!("Failed to access Android JVM: {}", e))?;
     let mut env = vm
@@ -772,7 +771,7 @@ fn resolve_android_content_uri_name(uri_str: &str) -> Result<String, String> {
 }
 
 #[cfg(target_os = "android")]
-fn read_android_content_uri(uri_str: &str) -> Result<Vec<u8>, String> {
+pub fn read_android_content_uri(uri_str: &str) -> Result<Vec<u8>, String> {
     let vm = unsafe { JavaVM::from_raw(android_context().vm().cast()) }
         .map_err(|e| format!("Failed to access Android JVM: {}", e))?;
     let mut env = vm
--- a/src-tauri/src/file_management.rs
+++ b/src-tauri/src/file_management.rs
@@ -714,106 +714,113 @@ fn parse_android_uri<'local>(
     Ok(uri)
 }
 
-#[cfg(target_os = "android")]
-fn resolve_android_content_uri_name(uri_str: &str) -> Result<String, String> {
-    let vm = unsafe { JavaVM::from_raw(android_context().vm().cast()) }
-        .map_err(|e| format!("Failed to access Android JVM: {}", e))?;
-    let mut env = vm
-        .attach_current_thread()
-        .map_err(|e| format!("Failed to attach current thread to Android JVM: {}", e))?;
+#[tauri::command]
+pub fn resolve_android_content_uri_name(uri_str: &str) -> Result<String, String> {
+    #[cfg(target_os = "android")]
+    {
+        let vm = unsafe { JavaVM::from_raw(android_context().vm().cast()) }
+            .map_err(|e| format!("Failed to access Android JVM: {}", e))?;
+        let mut env = vm
+            .attach_current_thread()
+            .map_err(|e| format!("Failed to attach current thread to Android JVM: {}", e))?;
 
-    let resolver = get_android_content_resolver(&mut env)?;
-    let uri = parse_android_uri(&mut env, uri_str)?;
-    let null_obj = JObject::null();
+        let resolver = get_android_content_resolver(&mut env)?;
+        let uri = parse_android_uri(&mut env, uri_str)?;
+        let null_obj = JObject::null();
 
-    let cursor = env
-        .call_method(
-            &resolver,
-            "query",
-            "(Landroid/net/Uri;[Ljava/lang/String;Ljava/lang/String;[Ljava/lang/String;Ljava/lang/String;)Landroid/database/Cursor;",
-            &[
-                (&uri).into(),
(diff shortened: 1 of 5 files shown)


CyberTimon:
Hey @waxm134, thanks for the PR!

Quick question regarding the Android file paths: I noticed we're using string splitting (.split('.') and rsplit_once('.')) to grab the extensions. I know this works great for standard file paths, but did you happen to test how this behaves with content:// URIs from the Storage Access Framework that don't have a file extension at the end?

What do you think about this?

waxm134:
The extension is part of the content URI string at least on my device. So just splitting until the first dot from the end of the string works there as well. I don't know whether this is the case for all Android variants. A cleaner approach would be to directly convert content URIs to fs paths using the tauri-android-fs plugin to not clutter the logic too much with android specific stuff. But this would be a larger refactoring of the code.

CyberTimon:
I double-checked the [Android docs](https://developer.android.com/training/secure-file-sharing/retrieve-info), and it turns out standard Android providers (like Google Drive) return opaque URIs where string splitting completely fails. The official docs specifically say we have to query the ContentResolver for OpenableColumns.DISPLAY_NAME to get the real filename and extension.

You don't even have to write this or add a new plugin - we actually already have a function for this in file_management.rs called resolve_android_content_uri_name.

waxm134:
Okay I see the issue. So for library input this should work by reusing the function mentioned. For the LUT it might get tricky on the backend side for the LUT caching. Because we cannot blindly call resolve_android_content_uri_name on the path passed because this path might not be accessible if this is not the first time the user picks the LUT file (the app is not allowed to access external files not explicitly picked by the user without external storage permission). Probably we can still call it and if it errors we know that the LUT file is already in the cache. I'll take a look.

waxm134:
With the new commit resolve_android_content_uri_name is called wherever we deal with content URIs

CyberTimon:
@waxm134 Thanks for implementing the fix so fast!
One quick request: please replace DefaultHasher for the .lut_cache filenames with a stable hash (like SHA256 or MD5) or base64 encoding. DefaultHasher is not guaranteed to be stable across different Rust compiler versions, which could lead to orphaned cache files on users' devices in future app updates. Also remove the duplicate #[cfg(target_os = "android")] in lut_processing.rs on line 189. Thanks!

waxm134:
@CyberTimon thanks for the review. Should be fixed with the new commit. I used blake3 for hashing as it is already used in file_management.rs.

CyberTimon:
@waxm134 
Thanks a lot! Since @Duecki1 [skipped the import dialogue](https://github.com/CyberTimon/RapidRAW/pull/1101), a merge conflict happened. As I haven't diged into this yet, do you have a quick solution to this? Duecki wants to completely skip the file import modal.

CyberTimon:
Fixed it myself :)

waxm134:
Hi @CyberTimon, thanks for merging! Just tested the latest main branch build and the Import button and the long press import is still there. What import dialog do you mean? Just noticed that the folder view is gone completely. With the new library folder users can in theory create sub-folders. Any thoughts on that?

Duecki1:
> Hi @CyberTimon, thanks for merging! Just tested the latest main branch build and the Import button and the long press import is still there. What import dialog do you mean? Just noticed that the folder view is gone completely. With the new library folder users can in theory create sub-folders. Any thoughts on that?

The import dialog meant is the pop up that came after selecting a raw file to import. It would ask you what to name it and where to place it. It's gone now 🫡

waxm134:
> The import dialog meant is the pop up that came after selecting a raw file to import. It would ask you what to name it and where to place it. It's gone now 🫡

Understood, I think it's a good idea to skip it.

Duecki1:
> > The import dialog meant is the pop up that came after selecting a raw file to import. It would ask you what to name it and where to place it. It's gone now 🫡
> 
> Understood, I think it's a good idea to skip it.

Okay so everyone who already has the android app, needs to click "Open Library" once instead of continue. How do we tell them? Or maybe force it in code?

waxm134:
> > > The import dialog meant is the pop up that came after selecting a raw file to import. It would ask you what to name it and where to place it. It's gone now 🫡
> > 
> > 
> > Understood, I think it's a good idea to skip it.
> 
> Okay so everyone who already has the android app, needs to click "Open Library" once instead of continue. How do we tell them? Or maybe force it in code?

What do you mean with that? Are pictures from the old FS path still shown in the app? Maybe that's a caching issue then. As the android build is marked as experimental for now, we should probably add a hint that we recommend backup of library data and clearing of the app data for new releases until the android port is considered stable.

Duecki1:
> > > > The import dialog meant is the pop up that came after selecting a raw file to import. It would ask you what to name it and where to place it. It's gone now 🫡
> > > 
> > > 
> > > Understood, I think it's a good idea to skip it.
> > 
> > 
> > Okay so everyone who already has the android app, needs to click "Open Library" once instead of continue. How do we tell them? Or maybe force it in code?
> 
> What do you mean with that? Are pictures from the old FS path still shown in the app? Maybe that's a caching issue then. As the android build is marked as experimental for now, we should probably add a hint that we recommend backup of library data and clearing of the app data for new releases until the android port is considered stable.

The problem is the "Continue" button, which doesnt refresh the already initiated path, which means it will use the old path. Only if the user clicks "other folder" it will refresh the path to the new one.

waxm134:
> The problem is the "Continue" button, which doesnt refresh the already initiated path, which means it will use the old path. Only if the user clicks "other folder" it will refresh the path to the new one.

I would not implement any workaround for that as the issue can practically not arise for anyone installing a non-experimental Android build in the future. Maybe @CyberTimon can comment on that. Probably a hint on how to backup old data and restore to the new path for a new release is sufficient.

PR #1113, opened 2026-04-30
Fix: clippy action warning

d1anarene:
CI's cargo clippy --all-targets --all-features -- -D warnings currently fails on main, blocking every open PR. 

image_processing.rs:1733 — needless_range_loop in the AgX tonemap LUT init
lib.rs:1958 — too_many_arguments (8/7) on process_image_for_export_pipeline after app_handle
lib.rs:2084 — same on process_image_for_export

Changes:
Switched the LUT loop to iter_mut().enumerate()
Added #[allow(clippy::too_many_arguments)] on both export functions, matching the existing pattern used elsewhere in lib.rs
No behavior chang


CyberTimon:
Thanks!

PR #1158, opened 2026-05-10
HEIF prototype via ffmpeg

fstanis:
https://github.com/CyberTimon/RapidRAW/pull/964#issuecomment-4415563126 for context

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -12,6 +12,7 @@ dependencies = [
  "bytemuck",
  "chrono",
  "fern",
+ "ffheif",
  "filetime",
  "futures",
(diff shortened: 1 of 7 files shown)


CyberTimon:
@fstanis It seems like the builds fail because you force C23 (`.clang_args(["-x", "c", "-std=c23"])`) in your [ffheif repository](https://github.com/fstanis/ffheif/blob/main/rust/ffheif-sys/build.rs). Is there a reason for this or can you edit it to use a widely supported one?

fstanis:
Right, that's just a remnant of some older C++ prototype - I changed it so that it uses `--std=c17` and rebuilt.

PR #1171, opened 2026-05-12
Recommended Vite Config for Tauri

RayJW:
## Description
Minor inconsistency between the recommendation of the [Tauri docs](https://tauri.app/start/frontend/vite/) and the Vite configuration. Stumbled upon this by accident and don't see an obvious reason this was done. From my understanding this should ensure, that Vite properly works with things like Tauris debug mode and slightly speed up dev builds.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made
- Use build config from Tauri docs

## Testing

**Test Configuration:**
- **OS:** CachyOS
- **Hardware:** Ryzen 9, AMD GPU

## Checklist

## AI Disclaimer:

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -24,4 +24,14 @@ export default defineConfig(async () => ({
       ignored: ["**/src-tauri/**"],
     },
   },
+
+  envPrefix: ['VITE_', 'TAURI_ENV_*'],
+  build: {
+    target:
+      process.env.TAURI_ENV_PLATFORM == 'windows'
+        ? 'chrome105'
+        : 'safari13',
+    minify: !process.env.TAURI_ENV_DEBUG ? 'esbuild' : false,
+    sourcemap: !!process.env.TAURI_ENV_DEBUG,
+  },
 }));


RayJW:
Hmm, I suspected something like this. I think these build failures might be cache related, because of the changed build process? I ran into something similar on my local system before clearing the build cache. Would you mind trying that @CyberTimon? Otherwise, I don't see, why this change could cause build failures.

CyberTimon:
Looks good to me. I will clear the cache soon :)

CyberTimon:
@RayJW I've cleared the cache but the workflow still fail. Do you have an idea why?

RayJW:
> @RayJW I've cleared the cache but the workflow still fail. Do you have an idea why?

Oh, I see the problem, the minified build requires `esbuild`, but it's not in the dependencies. Might've actually been my local cache causing the build to work. Can you quickly add `esbuild` as a dev dependency, that should fix the issue.

CyberTimon:
Still absolutely not building, should we remove this again?

CyberTimon:
Turns out Vite's built-in esbuild was just choking on the safari13 fallback from the Tauri docs because it's too old for some of our dependencies. I ended up removing the target block entirely so Vite uses its modern defaults, and everything should build again now.

RayJW:
> Turns out Vite's built-in esbuild was just choking on the safari13 fallback from the Tauri docs because it's too old for some of our dependencies. I ended up removing the target block entirely so Vite uses its modern defaults, and everything should build again now.

Yep, was just about to comment exactly the same haha :) Thanks for fixing this!

Issue #1160, opened 2026-05-10
BUG: Images a little blurry whe zoomed in


muddiver:
### What happened?

Affected are only the flatpak-versions: I shot a sequence of birds, and afterwards it seemed like all my images were dull, but the out-of-camera jpegs are sharp. When i open the CR3 in Darktable, it is sharp too. So i tried to change the previewmode, since i "only" have a 1920x1080 Display, i tried under setting->Processing to change the dynamic resolution instead of the fixed one. But no sharpness. Then i discovered, that when i toggle between the tabs "info" and "adjust" , the image is sharp for a moment, getting dull again when the toggle is finished. It is NOT the time the tool needs to load the CR3-file ( this is indicated by the waiting-circle beside the filename above the image). After the loading is done, the image gets sharper, but not as sharp as if i toggle between "info" and "adjust" Funny thing, when i change from adjust to inpaint, no sharpness, when in change from inpaint to adjust, short moment of sharpness.

### What did you expect to happen?

Sharp images when the camera did its job properly

### How can we reproduce this?

Simply load a CR3 file, zoom in to a point of interest and toggle between adjust and info.

### Operating System

Other Linux

### RapidRAW Version

1.5.5

### Additional information

Linux Mint, AMD APU 3200G

### Acknowledgement

- [x] I have verified that my camera and lenses are supported by the external libraries listed above, or my issue is unrelated to camera/lens support.
- [x] I have searched all existing issues, both open and closed, and did not find a duplicate.

muddiver:
Sorry, i forgot to mention that an exported image is as sharp as the In-Camera-Jpeg, its "only" the View in Rapidraw.

Duecki1:
Try using a higher Render Resolution Scale and/or enable high dpi rendering.

muddiver:
Hi Duecki,
I tried render resolution to 0,5, no success,
i cannot turn on High-DPI Rendering, because Rapidraw says i do not own a high.resolution screen, and Rapidraw is right ,-)
Setting Static preview to 3840Px no success, Rapidraw says the static preview will not affect the main editor preview..

Any other thoughts?  Can someone confirm the issue?
Greets, Tom

muddiver:
Just tested, it affects Jpegs too, not related to RAW..
And just to mention, i'm not pixelpeeping, i am at 100% view..

Duecki1:
[...]

PR #1198, opened 2026-05-22
fix: canvas renderer layer for live preview

rhysgs:
## Description

Solves issue relating to the live preview in some situations being put onto a different compositing/rendering layer resulting in significantly worse image quality in the live preview.

This specifically occurred when the image was zoomed in and then panned in such a way that the heavily scaled image had hidden overflow that would be put behind another div that was scrollable (in the case I found it was the editor side bar).

See here for my original evidence of the issue
https://github.com/CyberTimon/RapidRAW/issues/1160#issuecomment-4507646192

Admittedly I don't have a system I can use that would correctly allow the GPU and enable `isWgpuActive` for me to confirm this is doesn't break that.

At the very least please treat this as a POC that the blurry issue is real and can be solved.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made

Image is no longer plainly placed as a pannable image being scaled and moved. The view port previously being calculated is now instead used to render a crop of the original image into a canvas that occupies the live preview viewport. This simplifies the view layers greatly for the system having to render this.  The canvas is never behind other elements (compared to the image which was scaled and translated way out of bounds behind other things with hidden overflow).

Performance should be comperable I couldn't spot a difference by eye. I've only tested on my linux machine though.

Being in a canvas we now have full control of how the cropped viewport of the live preview is rendered.

## Screenshots/Videos

Before

[Screencast from 2026-05-22 22-04-29.webm](https://github.com/user-attachments/assets/c9a20393-72d0-4ced-aa88-24875fdfcc75)

After

[Screencast from 2026-05-22 22-02-18.webm](https://github.com/user-attachments/assets/1113bcbc-5ddf-4a4d-be2d-7677f22c66f4)

## Testing

**Test Configuration:**
- **OS:** Ubuntu 24.04
- **Hardware:** AMD 7600, AMD 7800XT, QHD Monitor

## Checklist

## Additional Notes

AI did a lot of the math for the translations to put the image into the canvas view. I'm sure this all could be refined further.

## AI Disclaimer:

### DIFF
--- a/src/components/panel/Editor.tsx
+++ b/src/components/panel/Editor.tsx
@@ -20,6 +20,8 @@ import { useSettingsStore } from '../../store/useSettingsStore';
 import { useUIStore } from '../../store/useUIStore';
 import { useLibraryStore } from '../../store/useLibraryStore';
 import { useAiMasking } from '../../hooks/useAiMasking';
+import { usePreviewCanvas } from './editor/usePreviewCanvas';
+import type { InteractivePatch } from '../../store/useEditorStore';
 
 const parseRgb = (rgbStr: string): [number, number, number, number] => {
   const match = rgbStr.match(/[\d.]+/g);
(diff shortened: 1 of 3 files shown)


CyberTimon:
Hi @rhysgs 

Thanks for the PR!

I noticed that the thumbnails no longer display at all while the actual image is loading. This happens universally, not just when the WGPU renderer is active. Could you please look into why this is happening and fix it?

Also, are you 100% sure the  performance isn't worse with this approach? I actually tried to implement a canvas-based preview here once before, but found that the performance (especially during rapid mouse movement) was much worse compared to just updating the CSS transforms on the SVG. Lastly, I think you should add an active/cleanup flag to your new Image() onload handlers, otherwise it can cause flickers while dragging sliders rapidly. :)

Thanks!

rhysgs:
@CyberTimon Thanks for checking out the PR, it is an interesting project! Some of the files are quite unwieldly (2,000 line react components) but it's quite exciting to see something that focuses on the basics (DarkTable is too in the weeds for me, feels more like a science experiment than a post production editor)

As for the performance concerns, this has a 100% chance of performing worse with this approach. There is no way it isn't with the extra layer of JS transforming and re-assigning the canvas content. It's not clear how to quantify that to measure the real impact but I found the experience on this branch perfectly acceptable. Though someone with less modern hardware would be a better fit for commenting on it. Modern hardware can brute force many things.

Unless the problem can be solved without a canvas respectfully it doesn't matter how fast the transformed svg strategy runs. If it doesn't fulfill the basic functional requirements of displaying an image without degredation it's not a solution. Software about editing images must render the live preview as faithfully as possible before anything else, optimisation before that is premature novelty.

It would be nice to keep the svg solution as it is simpler to reason about, but it appears to be a fundamental limitation of how layers are rendered when the composition layers includes a scrolling container. Tauri is nice that you don't need to bundle chromium, but that also means you don't control the version it runs on..

Thank you again for checking out the PR! I'll check out the onload hooks and flickering issue you've described that's no good, hard to spot regressions as someone very new to this project :)

PR #1207, opened 2026-05-24
fix gps metadata being stripped from exports (#1165)

### LINKED ISSUE #1165: BUG: Location data stripped from exported JPEG
### What happened?

If I open a JPEG file (note, not a RAW file) in RapidRaw, edit it and then export, location data is stripped from the exported image even though the setting to strip such data (in the export tab) is turned off.

I would expect that when the “strip location data” option is switched off, the location data would be preserved in the exported image.

Open a JPEG image in RapidRaw
Go to export tab
Ensure that the “remove location data” option is off
Export the file as a JPEG image
Open the exported file in an application that reads metadata eg Digikam
The location data is gone!

### Operating System

macOS Apple Silicon

### RapidRAW Version

1.5.5

### Additional information

I appreciate that RapdRaw is intended to edit RAW files and not JPEGs. However, it is possible to edit JPEGs too, and it is perverse that the setting to retain the location data is ignored when exporting.

### Acknowledgement

Waynting:
## Summary

Fixes #1165. GPS EXIF (location data) was always being stripped on export, even with **Remove location data** turned off.

## Root cause

`write_image_with_metadata` has three branches for sourcing the original image's metadata: the `.rrdata` sidecar, kamadak-exif (JPEG/PNG), and rawler (RAW). Only the kamadak and rawler branches had GPS-reading code.

Opening any image in the app triggers `persist_exif_if_missing` (`image_loader.rs:81`), which writes the `.rrdata` sidecar. From then on the sidecar branch fires on every export — silently bypassing both GPS code paths. So in practice `strip_gps=false` did nothing for any image you'd opened in the editor.

Affects JPEG (as reported) and RAW (same mechanism, just less likely to be noticed).

## Fix

GPS reading is orthogonal to the rest of the metadata sourcing. Lifted GPS reading out of the per-branch logic into two helpers and added a single tail call that runs unconditionally when `!strip_gps`:

```rust
if !strip_gps {
    if is_raw_file(original_path_str) {
        apply_gps_from_rawler(&mut metadata, original_path_str);
    } else {
        apply_gps_from_kamadak(&mut metadata, original_path);
    }
}
```

The sidecar / kamadak / rawler branches now handle only the non-GPS metadata they always did. GPS is re-read from the source file rather than parsed back from the sidecar's display strings (lossy — `"40 deg 26' 12.345\""` can't round-trip cleanly into the rationals `little_exif` needs).

## Behavior table

| | Before | After |
|---|---|---|
| JPEG + sidecar + strip_gps=false | ❌ stripped | ✅ preserved |
| JPEG + sidecar + strip_gps=true  | ✅ stripped | ✅ stripped |
| RAW  + sidecar + strip_gps=false | ❌ stripped | ✅ preserved |
| RAW  + sidecar + strip_gps=true  | ✅ stripped | ✅ stripped |
| TIFF output / `keep_metadata=false` | (early return) | (early return — unchanged) |

## Tests

Added two regression tests in `exif_processing::tests`. No new dependencies — `tempfile`, `image`, `little_exif`, `kamadak-exif` are all already in the project.

- `issue_1165_gps_preserved_when_sidecar_exists_and_strip_gps_false` — reproduces the exact bug trigger (real GPS tags + `.rrdata` sidecar via `persist_exif_if_missing`), asserts GPS lat/lon/alt are preserved in the exported bytes.
- `issue_1165_gps_stripped_when_strip_gps_true` — asserts the toggle still works.

Both pass:
```
test exif_processing::tests::issue_1165_gps_stripped_when_strip_gps_true ... ok
test exif_processing::tests::issue_1165_gps_preserved_when_sidecar_exists_and_strip_gps_false ... ok
```

Also verified:
- `cargo fmt --check` clean on the file
- `cargo clippy --lib --tests --no-deps` — no new warnings introduced

## Notes

- **Bonus**: the kamadak helper now also writes `GPSAltitudeRef` (above/below sea-level reference). The rawler branch already wrote it; added for symmetry. Harmless if downstream readers don't expect it.
- **Follow-up not done here**: `get_string_val` closure is now defined in two places (the helper + the inline kamadak branch). Kept this PR narrow; happy to lift it to a module-level helper in a follow-up if you'd like.

## Test plan

- [x] `cargo test --lib exif_processing::tests`
- [x] `cargo fmt -- --check src/exif_processing.rs`
- [x] `cargo clippy --lib --tests --no-deps`
- [ ] Manual: open a GPS-tagged JPEG, export with "Remove location data" off → `exiftool -GPS:all exported.jpg` shows coords
- [ ] Manual: same with the toggle on → no GPS tags in output

### DIFF
--- a/src-tauri/src/exif_processing.rs
+++ b/src-tauri/src/exif_processing.rs
@@ -610,6 +610,138 @@ pub fn get_creation_date_from_bytes(path_hint: &str, file_bytes: &[u8]) -> DateT
     Utc::now()
 }
 
+fn apply_gps_from_kamadak(metadata: &mut Metadata, original_path: &Path) {
+    let Ok(file) = std::fs::File::open(original_path) else {
+        return;
+    };
+    let mut bufreader = std::io::BufReader::new(&file);
+    let exifreader = exif::Reader::new();
+    let Ok(exif_obj) = exifreader.read_from_container(&mut bufreader) else {
+        return;
+    };
+
+    let get_string_val = |field: &exif::Field| -> String {
+        match &field.value {
+            exif::Value::Ascii(vec) => vec
+                .iter()
+                .map(|v| {
+                    String::from_utf8_lossy(v)
+                        .trim_matches(char::from(0))
+                        .to_string()
+                })
+                .collect::<Vec<String>>()
+                .join(" "),
+            _ => field
+                .display_value()
+                .to_string()
+                .replace("\"", "")
+                .trim()
+                .to_string(),
+        }
+    };
+
+    if let Some(f) = exif_obj.get_field(exif::Tag::GPSLatitude, exif::In::PRIMARY)
+        && let exif::Value::Rational(v) = &f.value
+        && v.len() >= 3
+    {
+        metadata.set_tag(ExifTag::GPSLatitude(vec![
+            to_ur64(&v[0]),
+            to_ur64(&v[1]),
+            to_ur64(&v[2]),
+        ]));
+    }
+    if let Some(f) = exif_obj.get_field(exif::Tag::GPSLatitudeRef, exif::In::PRIMARY) {
+        metadata.set_tag(ExifTag::GPSLatitudeRef(get_string_val(f)));
+    }
+    if let Some(f) = exif_obj.get_field(exif::Tag::GPSLongitude, exif::In::PRIMARY)
+        && let exif::Value::Rational(v) = &f.value
+        && v.len() >= 3
+    {
+        metadata.set_tag(ExifTag::GPSLongitude(vec![
+            to_ur64(&v[0]),
+            to_ur64(&v[1]),
+            to_ur64(&v[2]),
+        ]));
+    }
+    if let Some(f) = exif_obj.get_field(exif::Tag::GPSLongitudeRef, exif::In::PRIMARY) {
+        metadata.set_tag(ExifTag::GPSLongitudeRef(get_string_val(f)));
+    }
+    if let Some(f) = exif_obj.get_field(exif::Tag::GPSAltitude, exif::In::PRIMARY)
+        && let exif::Value::Rational(v) = &f.value
+        && !v.is_empty()
+    {
+        metadata.set_tag(ExifTag::GPSAltitude(vec![to_ur64(&v[0])]));
+    }
+    if let Some(f) = exif_obj.get_field(exif::Tag::GPSAltitudeRef, exif::In::PRIMARY)
+        && let Some(val) = f.value.get_uint(0)
+    {
+        metadata.set_tag(ExifTag::GPSAltitudeRef(vec![val as u8]));
+    }
+}
+
+fn apply_gps_from_rawler(metadata: &mut Metadata, original_path_str: &str) {
+    let loader = rawler::RawLoader::new();
+    let Ok(raw_source) = rawler::rawsource::RawSource::new(Path::new(original_path_str)) else {
+        return;
+    };
+    let Ok(decoder) = loader.get_decoder(&raw_source) else {
+        return;
+    };
+    let Ok(meta) = decoder.raw_metadata(&raw_source, &Default::default()) else {
+        return;
+    };
+    let Some(gps) = meta.exif.gps else {
+        return;
+    };
+    if let Some(lat) = gps.gps_latitude {
+        metadata.set_tag(ExifTag::GPSLatitude(vec![
+            uR64 {
+                nominator: lat[0].n,
+                denominator: lat[0].d,
+            },
+            uR64 {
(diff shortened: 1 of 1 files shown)


Wakaba0815:
Can confirm this independently on v1.5.8 (Linux) with a real-world case — might be a useful additional data point since it shows the sidecar-path trigger very clearly.

Source: JPEG straight from a Samsung Galaxy S22 Ultra (firmware S908BXXSNGZD7), synced via Nextcloud instant upload. Full GPS EXIF block present and verified in the original:

[GPS]           GPSVersionID                    : 2.2.0.0
[GPS]           GPSLatitudeRef                  : North
[GPS]           GPSLatitude                     : XX deg XX' XX.XX"
[GPS]           GPSLongitudeRef                 : East
[GPS]           GPSLongitude                    : XX deg XX' XX.XX"
[GPS]           GPSAltitudeRef                  : Above Sea Level
[GPS]           GPSAltitude                     : XXX m
[GPS]           GPSDateStamp                    : 2026:07:01
[GPS]           GPSTimeStamp                    : 11:40:10
[GPS]           GPSProcessingMethod             : GPS

(via exiftool -a -G1 -s file.jpg | grep -i gps — note: plain exiftool -gps:all without -a misleadingly shows only partial ref tags on this file due to duplicate IFD blocks, worth keeping in mind if others report inconsistent repro results.)

After opening the file in RapidRAW and exporting (tested with "Strip GPS" both ON and OFF — no difference in outcome), the exported JPEG has zero GPS tags, not even the refs:

exiftool -a -G1 -s export.jpg | grep -i gps
(no output)

This lines up exactly with the root cause described in #1207 — once the image is opened in the editor, persist_exif_if_missing writes a .rrdata sidecar, and from that point on write_image_with_metadata takes the sidecar path for export, which has no GPS read logic at all (unlike the kamadak-exif / rawler paths). So the "Strip GPS" toggle has no effect once a sidecar exists — it's not being ignored, it's just not reached.

Confirms this isn't camera- or format-specific — happy to test the fix in #1207 against this file if it'd help verification before merge.

Waynting:
Thanks @Wakaba0815 — this is a really useful data point, especially the note about needing `exiftool -a` on files with duplicate IFD blocks; that's exactly the kind of real-world case the synthetic test files in this PR can't reproduce.

Your read of the root cause matches the fix: GPS reading is now lifted out of the three metadata-source branches (sidecar / kamadak-exif / rawler) and runs unconditionally at the end of `write_image_with_metadata` whenever Strip GPS is off — so the sidecar path can no longer bypass it, and the toggle works regardless of whether an `.rrdata` sidecar exists.

Testing against your S22 Ultra file before merge would be very welcome. To build this branch:

```bash
gh pr checkout 1207    # or: git clone -b fix/1165-gps-export-preserved https://github.com/Waynting/RapidRAW
npm install
npm start              # runs the tauri dev build
```

Then repeat your exact repro: open the file in the editor (so the `.rrdata` sidecar exists), export once with Strip GPS **off** and once with it **on**, and verify with `exiftool -a -G1 -s`. Expected: full GPS block preserved (including refs and altitude) with the toggle off, zero GPS tags with it on.

If you want to run the regression tests too: `cd src-tauri && cargo test exif_processing` — the first one reproduces the sidecar trigger you described.

Wakaba0815:
Tested PR #1207 against my S22 Ultra file (built from `fix/1165-gps-export-preserved`, 
`npm start` dev build on Linux).

Repro steps: opened the file in the editor (so the `.rrdata` sidecar exists, matching 
my original report), then exported once with Strip GPS **off** and once **on**.

**Strip GPS off** — full GPS block preserved:
[GPS]           GPSLatitudeRef                  : North
[GPS]           GPSLatitude                     : XX deg XX' XX.XX"
[GPS]           GPSLongitudeRef                 : East
[GPS]           GPSLongitude                    : XX deg XX' XX.XX"
[GPS]           GPSAltitudeRef                  : Above Sea Level
[GPS]           GPSAltitude                     : XXX m
[Composite]     GPSAltitude                     : XXX m Above Sea Level
[Composite]     GPSLatitude                     : XX deg XX' XX.XX" N
[Composite]     GPSLongitude                    : XX deg XX' XX.XX" E
[Composite]     GPSPosition                     : XX deg XX' XX.XX" N, XX deg XX' XX.XX" E
**Strip GPS on** — zero GPS tags, as expected.

Confirmed via `exiftool -a -G1 -s <file> | grep -i gps` on both exports. Fix works 
correctly for my case — the sidecar-triggered bypass is resolved. 👍

CyberTimon:
Thanks. Implemented in https://github.com/CyberTimon/RapidRAW/commit/e801cd3b31d70c07e1ed2865f7190ca55a43aee7, without the test suite :)

PR #1307, opened 2026-06-26
Backend robustness fixes + module/component refactors (from a code review)

SandeepSubba:
This branch came out of a code-analysis pass over RapidRAW. Three commits,
each scoped to one concern.

### 1. Backend correctness & robustness (`fix:`)
- **Stale-cache fix (the notable one):** `calculate_transform_hash` hashed
  each AI patch by base64 *string length* instead of content, so two
  different patch results of identical dimensions could collide and leave
  the old patch composited. Now hashes the actual content.
- **AI model output validation:** the SAM decoder, sky-seg, U-2-Netp,
  Depth-Anything, and LaMa paths indexed model output assuming an exact
  tensor shape and `.unwrap()`-ed. A model whose output resolution differs
  now returns a clean error instead of panicking.
- **Export guards:** bounds-check the frontend `mask_index`; guard
  `calculate_resize_target` against zero-dimension images (was Inf/NaN).
- **lut_cache cap:** it was keyed by path and never cleared by the
  session/image resets, so it grew unbounded — now capped at 32 entries.

### 2. Extract SettingsPanel widgets (`refactor:`)
Move 7 self-contained presentational components into
`components/panel/settings/widgets.tsx`. `SettingsPanel.tsx` 2415 → 2032 lines.

### 3. Split two oversized Rust files (`refactor:`)
Peel cohesive clusters into submodules (`use super::*` + `pub use`, so the
command paths in `lib.rs` are unchanged):
- `file_management/albums.rs`, `file_management/folder_tree.rs`
- `image_processing/analysis.rs`

`file_management.rs` 3661 → 3135, `image_processing.rs` 3262 → 2555.

### Verification
- `cargo check` — clean (no errors/warnings).
- `npm run typecheck` — clean.

These are deliberately small, self-contained changes; happy to split into
separate PRs or adjust anything if it doesn't fit your direction.


akhater:
Good catch on the cache key. Two patches of the same base64 length shared an entry — and base64 of a fixed-size region is very often the same length, so it bites more than it looks.

Confirmed and covered with regression tests in [Argentum](https://github.com/akhater/Argentum) — our RapidRAW fork, and as sharp as a prime: [`79a0658`](https://github.com/akhater/Argentum/commit/79a06584). Marked so we drop our copy when this lands.

Thanks — wrong picture, no error, nothing to grep for. Nasty one.

PR #1324, opened 2026-07-03
Fall back to embedded previews for undecodable raws

SandeepSubba:
## Problem
Some raw files fail to load entirely — no thumbnail, unopenable in the editor:
- **DNG 1.7 with 8-bit lossy JPEG compression** (increasingly common from smartphones/Adobe lossy exports) — rawler's lossless-JPEG decoder rejects them with `ljpeg: sof.precision 8`, surfaced as "Failed to decode image, possibly corrupt image"
- **Raws with stripped make/model EXIF** (anonymized/re-exported files) — rawler gets "No decoder found"

## Fix
When raw development fails, fall back to the file's embedded JPEG preview instead of erroring:
- Walk the TIFF IFD tree directly for the **largest** embedded JPEG (new-style strips with reduced-resolution subfile type, plus old-style `JPEGInterchangeFormat` tags). This is decoder-independent, and picks the full-size preview — rawler's own extractor returns the small thumbnail for DNG 1.7 files.
- Apply EXIF orientation, linearize (the raw pipeline tone-maps linear input), and continue as an 8-bit image.
- A warning is logged when the fallback engages; rawler's extractor remains the fallback for non-TIFF layouts (e.g. RAF).

The file becomes viewable and editable rather than a hard failure. Edits obviously lack raw latitude (it's the 8-bit preview), but that beats a broken file.

## Verification
Tested against a DNG 1.7 sample (5464×8192, lossy 8-bit): previously failed with the decode error; now opens at the full embedded-preview resolution with correct orientation. Regular raws are unaffected (fallback only runs after development fails).

### DIFF
--- a/src-tauri/src/image_loader.rs
+++ b/src-tauri/src/image_loader.rs
@@ -5,7 +5,9 @@ use crate::exif_processing;
 use crate::file_management::{parse_virtual_path, read_file_mapped};
 use crate::formats::is_raw_file;
 use crate::image_processing::ImageMetadata;
-use crate::image_processing::{apply_orientation, remove_raw_artifacts_and_enhance};
+use crate::image_processing::{
+    apply_orientation, apply_srgb_to_linear, remove_raw_artifacts_and_enhance,
+};
 use crate::mask_generation::{MaskDefinition, SubMask, generate_mask_bitmap};
 use crate::raw_processing::develop_raw_image;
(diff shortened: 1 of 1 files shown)


CyberTimon:
LGTM! I changed some stuff so it doesn't read the file twice. Thanks for the PR :)

PR #1334, opened 2026-07-05
Versions (virtual copies): multiple saved looks per photo

SandeepSubba:
## What

Adds **Versions** — named snapshots of the complete edit state, stored inside the sidecar — so one RAW can carry several looks ("client picks", color vs. B&W, subtle vs. bold retouch) and switch between them with one click.

Version switching

## How it works

A **Versions** section at the top of the Presets panel:

- **+** saves the current edit as a new version (full snapshot: adjustments, masks, AI edits, crop)
- **click** a version to apply it
- right-click for **Apply / Overwrite with Current Edit / Rename / Delete**

Version context menu

## Design notes

- **No backend schema change.** Versions live in `adjustments.versions[]` (`{id, name, createdAt, state}`); the sidecar already stores adjustments as raw JSON, so they round-trip untouched.
- **Undoable + auto-saved.** Every version operation goes through `setAdjustments`, so it participates in history and the normal save path like any other edit.
- **Isolated.** Versions are excluded from copy/paste and multi-select sync (they're not in the copyable-key whitelist), so pasting adjustments to other photos never drags snapshots along.
- **Reset-safe.** `reset_adjustments_for_paths` now preserves the `versions` key — clearing the working edit can't destroy saved looks.

## Testing

- Full round-trip verified in-app: save two versions with different looks, switch between them repeatedly, restart-persistence via sidecar confirmed; version operations undo cleanly
- `cargo check`, `cargo fmt --check` clean; `tsc` introduces no new errors

### DIFF
--- a/src-tauri/src/file_management.rs
+++ b/src-tauri/src/file_management.rs
@@ -2260,7 +2260,16 @@ pub async fn reset_adjustments_for_paths(
 
             let mut existing_metadata = crate::exif_processing::load_sidecar(&sidecar_path);
 
+            // Saved versions ("virtual copies") survive a reset of the working state.
+            let preserved_versions = existing_metadata
+                .adjustments
+                .get("versions")
+                .filter(|v| v.as_array().is_some_and(|a| !a.is_empty()))
+                .cloned();
             existing_metadata.adjustments = serde_json::json!({});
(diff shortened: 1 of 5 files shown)


SandeepSubba:
Withdrawing this one.

PR #1349, opened 2026-07-07
Film negative conversion: non-destructive & in-library (replaces the modal/TIFF flow)

SandeepSubba:
## Summary

Reworks the Film Negative Converter from a modal that bakes a separate `_Positive.tiff` into a **non-destructive, in-library** conversion that fits RapidRAW's sidecar model.

You right-click a negative → **Convert Negative**. It inverts to a neutral positive *in place* (no new file, no duplicate library entry), and you do all the tuning — exposure, white balance, contrast, curves, masks — in the normal Develop module on top. Right-click → **Revert to Negative** turns it back off.

## Why

The old flow was self-contained because it was destructive: once you exported the TIFF there was no going back, so the modal carried its own exposure / contrast / white-balance sliders. Those duplicate what Develop already does. The only genuinely negative-specific step is the density inversion + per-channel auto-bounds — everything else belongs in Develop. Making it non-destructive removes the duplication and the duplicate files, and makes the conversion re-tunable and reversible.

## What changed

**Behaviour**
- Right-click **Convert Negative** / **Revert to Negative** in both the editor and the library/filmstrip context menus. The label toggles based on whether the image(s) are already converted; batch selection works.
- The conversion (density inversion + per-channel auto-bounds) is stored in the image's `.rrdata` sidecar as `negativeConversion: { enabled, bounds }` and applied on load, so the editor, thumbnails and export all render the positive. The original raw is never touched.
- The Convert Negative modal is removed.

**Backend**
- Replaces `preview_negative_conversion` / `convert_negatives` with a single `set_negative_conversion(paths, enabled)`.
- `maybe_apply_negative` inverts a flagged negative at the single decode chokepoint (`load_base_image_from_bytes`), so every consumer agrees.
- Auto-bounds are still computed at convert time (same NegPy-derived analysis) for a good neutral starting point.
- Adjustment saves preserve the backend-owned `negativeConversion` block, so a normal edit or navigation can't strip it; thumbnails regenerate on convert/revert.

**Notes**
- Backward compatible — existing sidecar conversions still render.
- The inversion math is unchanged (still the NegPy-derived pipeline, GPL-3.0, credited as before).


CyberTimon:
Great PR! 

My notes from a quick testing:
 - Batch negative conversion works but batch reverting doesn't do anything
 - Too many AI generated code comments. Please remove these code comments.

Thanks a lot, looks good otherwise :)

evilbuck:
Is this dead? I would love to see this drop.

SandeepSubba:
> Is this dead? I would love to see this drop.

It's obvious, mine and Timon's vision doesn't match so I have made my own fork with the several features that I need. I have made image importer with culling. I have also made film scanning feature. I have Plustek Film scanner, so it works with it. For other scanners, I don't have it to test it out and develop the feature.

You can find it in my github repo
https://github.com/SandeepSubba/RapidRAW

Disclaimer: It is fully made with Claude Code

CyberTimon:
It's absolutely not dead yet and I would love to implement something like this into RapidRAW. The reason this is on hold is because I want to apply the same to denoising and other functions to keep it consistent with the tools. I haven't found a elegant solution yet, as this can be done in under 1 second (which is fine for image loading), while denoising takes several minutes..

PR #1350, opened 2026-07-08
Native File Transfer for Thumbnails

RayJW:
## Description
This gets rid of the Base64 file transfer for the thumbnails and instead uses a native path. While the savings are miniscule we still save 33% overhead in transferred bits and the CPU cycles for encoding and decoding the data. In my testing I saw savings in RAM usage in the range of tens of MBs. While the impact is relatively small I still think this change is worth it in the name of efficiency, since it doesn't really introduce additional complexity.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Performance improvement
- [X] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made

- Allow Tauri to access the thumbnail cache location
- Enable `protocol-asset` feature for Tauri
- Pass around thumbnails by path instead of Base64 string
- Refactor some duplicated code into a function

## Testing

**Test Configuration:**
- **OS:** CachyOS
- **Hardware:** Ryzen 7, AMD Graphics, 32 GB RAM

## Checklist

## AI Disclaimer:
- [X] This PR is AI-generated but guided by a human

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -2557,6 +2557,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "http-range"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "21dec9db110f5f872ed9699c3ecf50cf16f423502706ba5c72462e28d3157573"
+
+[[package]]
 name = "httparse"
 version = "1.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -6353,6 +6359,7 @@ dependencies = [
  "gtk",
  "heck 0.5.0",
  "http",
+ "http-range",
  "jni 0.21.1",
  "libc",
  "log",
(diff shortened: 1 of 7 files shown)


CyberTimon:
Great PR! Looks good to me :)

R-Laine:
I believe this commit is what has caused thumbnail loading to become broken in some scenarios on windows, my report #1403 though if I'm wrong do call me out

You changed thumbnail loading from base64 data URLs to Tauri's native file transfer via the asset protocol. The asset protocol configuration $APPCACHE/thumbnails/* can fail to resolve correctly on Windows which could be the root cause of my issue

RayJW:
> I believe this commit is what has caused thumbnail loading to become broken in some scenarios on windows, my report #1403 though if I'm wrong do call me out
> 
> You changed thumbnail loading from base64 data URLs to Tauri's native file transfer via the asset protocol. The asset protocol configuration $APPCACHE/thumbnails/* can fail to resolve correctly on Windows which could be the root cause of my issue

I think you might be right, thanks for tracking this down!

Would you mind testing #1410 to confirm the patch I wrote actually resolves the issue on Windows?

R-Laine:
Thumbnails do load with the fix

Issue #1322, opened 2026-07-02
FEATURE: Export JPEG XL & WepB with EXIF metadata


leetskills:
### Feature Description

JPEG export has a toggle for 'Save with Metadata' (and 'Remove GPS data'), but these options aren't available with JPEG XL or WebP. I don't want to be using JPEGs in 2026.

### Similar feature requests

- [x] I have searched existing issues and discussions and didn't find a duplicate

Issue #503, opened 2025-12-04
Exif metadata not present in images exported


metboef:
### What happened?

I tried both keeping or not the original metadata, but no Exif data is shown in the jpg exported images (as checked in Fastone, or Hugin, or RapidRaw itself)

### What did you expect to happen?

to get the exif metadata in exported images

### How can we reproduce this?

export an image

### Operating System

Windows 10

### RapidRAW Version

1.4.6

### Similar issues

- [x] I have searched existing issues and didn't find a duplicate

CyberTimon:
Please look for exisiting issues first:
 - https://github.com/CyberTimon/RapidRAW/issues/416
 - https://github.com/CyberTimon/RapidRAW/issues/401
 - https://github.com/CyberTimon/RapidRAW/discussions/101
 - https://github.com/CyberTimon/RapidRAW/issues/40

Issue #427, opened 2025-11-01
Metadata not exported


Bagel94:
### What happened?

When exporting photos, metadata is not being exported with the file. So, when adding photos to external albums. Date filtering does not work as it's not in the metadata

### What did you expect to happen?

To have metadata exported 

### Operating System

Windows 11

### RapidRAW Version

Newest

### Similar issues

- [x] I have searched existing issues and didn't find a duplicate

CyberTimon:
Thank you for the report and for the feedback. It's a duplicate of [this discussion](https://github.com/CyberTimon/RapidRAW/discussions/101) and as I don't maintain the EXIF writing crate and EXIF support in general is pretty limited in Rust (the underlying programming language) I can't really give a ETA or a fix.

Bagel94:
Sorry for the duplicate, must have missed the other when I was trying to find one related. Thanks for the response though!

CyberTimon:
I was able to implement it correctly now: https://github.com/CyberTimon/RapidRAW/commit/95d76822e4f245a9810e23bd52258b0870a4d588

It works now for most of the tags. Will appear in the next release 💯

Issue #401, opened 2025-10-20
Metadata not preserved in exported files


louloutyro:
### What happened?

Whenever I export .NEF files to JPEG, whatever "metadata" settings I use in the export menu, all the metadata is erased, even the capture date.

Loving the software by the way, thank you for everything.

<img width="593" height="346" alt="Image" src="https://github.com/user-attachments/assets/b95189c2-474e-40e7-aa5d-e643e9c31db6" />
<img width="1512" height="948" alt="Image" src="https://github.com/user-attachments/assets/a2e2b058-4b60-49f0-83d3-93f8138b6d49" />

### What did you expect to happen?

Expected the metadata to be written into the JPEG file.

### How can we reproduce this?

1. Import a raw file with metadata.
2. Export it with "keep original metadata" activated.
3. See that the original metadata isn’t preserved.

### Operating System

macOS Apple Silicon

### RapidRAW Version

v1.4.3

### Additional information

Tried every "metadata" setting in the export menu. Everything produces the same results.

<img width="305" height="125" alt="Image" src="https://github.com/user-attachments/assets/ffcbbc9a-750e-43ea-913a-7c62f03484ab" />

### Similar issues

- [x] I have searched existing issues and didn't find a duplicate

CyberTimon:
Thank you for the report and for the feedback. It's a duplicate of [this discussion](https://github.com/CyberTimon/RapidRAW/discussions/101) and as I don't maintain the EXIF writing crate and EXIF support in general is pretty limited in Rust (the underlying programming language) I can't really give a ETA or a fix.

PR #1373, opened 2026-07-13
fix: preserve the full EXIF block in exports instead of the sidecar subset (#1165)

### LINKED ISSUE #1165: BUG: Location data stripped from exported JPEG

Serraniel:
## Description
Exporting with **Save with Metadata** enabled silently drops most of the EXIF data. From a Fujifilm X-T50 JPEG with 78 EXIF tags, only 15 survive the export: the whole GPS block, `DateTime`/`OffsetTime*`/`SubSecTime*`, `Flash`, `ExposureBiasValue`, `ApertureValue`, `ShutterSpeedValue`, the MakerNote, body/lens serial numbers, `LensSpecification`, the Interop IFD and more are gone.

**Root cause:** `write_image_with_metadata` sources tags from three branches (`.rrdata` sidecar → kamadak-exif → rawler). Opening an image triggers `persist_exif_if_missing`, which creates the sidecar — so from then on the sidecar branch always wins. The sidecar only stores a display-formatted subset of the tags (e.g. `"ApertureValue": "7.4 EV"`, MakerNote as a truncated hex dump), so it cannot reconstruct the original EXIF, and the kamadak/rawler branches (including their GPS handling) are never reached.

**Fix:** for non-RAW sources, copy the complete EXIF block (IFD0 + Exif/GPS/Interop sub-IFDs, including MakerNote and unknown tags) from the source file via `little_exif`, instead of reconstructing a subset:

- The sidecar is still consulted afterwards, but only for the fields that are editable in the metadata panel (`Artist`, `Copyright`, `ImageDescription`, `UserComment`) — user edits override the source values, and a cleared field stays cleared in the export.
- `strip_gps=true` still removes the entire GPS IFD (all `GPS*` tags, not just lat/lon/alt).
- `Orientation` is reset to 1 as before (rotation is baked into the exported pixels), and `ExifImageWidth`/`ExifImageHeight` are now set to the actual export dimensions instead of carrying the stale source values.
- IFD1 is not copied, since it contains the original's embedded thumbnail which no longer matches the exported pixels.
- RAW sources are unchanged: they keep the existing sidecar/rawler reconstruction (a full block copy is not possible there, since `little_exif` cannot parse RAW containers).

Fixes #1165 (the GPS part of it as a side effect, see below). The closed duplicates #503, #427 and #401 report the same symptom.

Relationship to other PRs:
- **#1207** fixes the GPS subset of the same root cause by unconditionally re-reading GPS from the source file. This PR is complementary: it covers *all* tags for non-RAW sources, while #1207 additionally restores GPS for RAW-with-sidecar exports, which this PR intentionally leaves untouched. Both can land: whichever merges second just needs a trivial rebase of the (textually adjacent) changes in `exif_processing.rs`. The regression tests here follow the pattern introduced in #1207.
- **#1246** re-attaches `OffsetTime*` to JPEG exports; the export half of that PR is covered by the full-block copy for non-RAW sources (its `get_creation_date_from_path` timezone fix is independent and still valuable).

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made
- `exif_processing::write_image_with_metadata`: copy the full EXIF block from non-RAW source files (new `copy_full_exif_from_source`), skipping the GPS IFD when `strip_gps` is set and skipping IFD1; fall back to the previous sidecar/kamadak/rawler branches when the source has no parseable EXIF or is a RAW file.
- New `apply_sidecar_field_overrides`: the four panel-editable fields are applied from the sidecar on top of the copied block.
- Set `ExifImageWidth`/`ExifImageHeight` to the actual dimensions of the exported image.
- Regression tests: full tag preservation (incl. MakerNote byte-identity) with a sidecar present, complete GPS removal with `strip_gps=true`, sidecar edits overriding/clearing source values, orientation reset + dimension tag update.
- Separate commit for the WebP half of #1322: `little_exif` converts the exporter's simple-format WebP to the extended format and attaches an EXIF chunk, so WebP EXIF only needed a file-type mapping (tested with both the lossy VP8 output of the export pipeline and lossless VP8L).

## Screenshots/Videos
Before/after on a real Fujifilm X-T50 JPEG, exported with *Save with Metadata* ON and *Remove GPS Data* OFF after the image had been opened once (sidecar present). Tag counts from `exiftool -a -G1 -s`, excluding the ExifTool/System/File/Composite/JFIF pseudo-groups:

| | tags | missing vs. original |
|---|---|---|
| original DSCF0563.JPG | 155 | — |
| RapidRAW 1.5.8 export | **14** | GPS block, MakerNote (all 70+ Fujifilm tags), DateTime, OffsetTime\*, SubSecTime\*, Flash, ApertureValue, ShutterSpeedValue, ExposureBiasValue, body/lens serials, Interop IFD, PrintIM, ... |
| this PR | **141** | only IFD1 (stale embedded thumbnail, intentionally dropped), APP2 FlashPix preview + XMP Rating (separate segments, also dropped by 1.5.8, out of scope) |

A tag-by-tag value diff of original vs. new export shows exactly the intended differences: `Orientation` (reset, rotation baked in), `Software` (RapidRAW), `ExifImageWidth/Height` (now the real export dimensions). With *Remove GPS Data* ON the export contains zero `GPS*` tags. exiftool parses the copied Fujifilm MakerNote in the export without errors, so the offsets survive the copy.

## Testing

`cargo fmt --check`, `cargo clippy --lib --tests --no-deps` and `cargo test` pass. Verified against a real Fujifilm X-T50 JPEG (see table above): tag-by-tag `exiftool` diff of original vs. export matches except for the intentional differences (`Orientation`, pixel dimensions, `Software`, thumbnail).

**Test Configuration:**
- **OS:** Arch Linux (RapidRAW 1.5.8 via rapidraw-bin AUR for the "before" state)
- **Hardware:** AMD Ryzen 7 5700X, Radeon RX 6800 XT

## Checklist

## Additional Notes
- MakerNote is copied as an opaque blob. For self-contained MakerNotes (e.g. Fujifilm, whose offsets are relative to the MakerNote itself) this round-trips perfectly; MakerNotes that use absolute file offsets may have dangling offsets after the copy — still strictly better than dropping the MakerNote entirely, and the same trade-off most non-exiftool writers make.
- `exiftool -validate` reports "entries out of order" notices on the exports because `little_exif` appends the ExifOffset/GPSInfo/InteropOffset pointer tags after the (otherwise sorted) IFD entries at encode time. That's a pre-existing quirk of the `little_exif` encoder (it just wasn't visible before because so few tags survived), readers are unaffected, and it would be fixed upstream, not here.
- RAW-with-sidecar exports still only get the reconstructed subset (plus GPS once #1207 lands). A follow-up could extract the full EXIF from the embedded preview JPEG.

## AI Disclaimer:

### DIFF
--- a/src-tauri/src/exif_processing.rs
+++ b/src-tauri/src/exif_processing.rs
@@ -783,6 +783,7 @@ pub fn write_image_with_metadata(
             as_zTXt_chunk: true,
         },
         "tiff" => FileExtension::TIFF,
+        "webp" => FileExtension::WEBP,
         _ => return Ok(()),
     };
 
@@ -1595,5 +1596,50 @@ mod tests {
             (Some(32), Some(64)),
             "dimension tags must match the exported pixels"
         );
+    }
+
+    #[test]
+    fn issue_1322_webp_export_carries_exif() {
+        let dir = TempDir::new().unwrap();
+        let (jpeg_path, _) = setup_jpeg_with_sidecar(&dir);
+
+        let img: ImageBuffer<Rgb<u8>, Vec<u8>> =
+            ImageBuffer::from_pixel(64, 64, Rgb([200, 100, 50]));
+        let mut lossless_bytes = Vec::new();
+        img.write_to(
+            &mut Cursor::new(&mut lossless_bytes),
+            image::ImageFormat::WebP,
+        )
+        .unwrap();
+        // Same encoder the export pipeline uses (lossy VP8).
+        let lossy_bytes = webp::Encoder::from_image(&image::DynamicImage::ImageRgb8(img))
+            .unwrap()
+            .encode(80.0)
+            .to_vec();
+
+        for mut export_bytes in [lossless_bytes, lossy_bytes] {
+            write_image_with_metadata(
+                &mut export_bytes,
+                jpeg_path.to_str().unwrap(),
+                "webp",
+                true,
+                false,
+            )
+            .unwrap();
+
+            let exif_obj = read_export_exif(&export_bytes);
+            for tag in [
+                exif::Tag::Make,
+                exif::Tag::Flash,
+                exif::Tag::MakerNote,
+                exif::Tag::GPSLatitude,
+            ] {
+                assert!(
+                    exif_obj.get_field(tag, In::PRIMARY).is_some(),
+                    "{} must be written to the WebP EXIF chunk (issue #1322)",
+                    tag
+                );
+            }
+        }
     }
 }
--- a/src-tauri/src/exif_processing.rs
+++ b/src-tauri/src/exif_processing.rs
@@ -9,6 +9,7 @@ use chrono::{DateTime, NaiveDateTime, Utc};
 use exif::{Exif, In, Value};
 use little_exif::exif_tag::ExifTag;
 use little_exif::filetype::FileExtension;
+use little_exif::ifd::ExifTagGroup;
 use little_exif::metadata::Metadata;
 use little_exif::rational::{iR64, uR64};
 use rawler::decoders::RawMetadata;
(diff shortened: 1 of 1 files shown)


CyberTimon:
Thanks for the PR!
Before I review it, remove all code comments and tests.

r3nor:
I think this feature is important for JPEGXL and other formats, since currently they do not support metadata in the file...

flloschy:
Whats the rough ETA on merging this? I just screwed up half my immich timeline by uploading ~400 edited photos, because Timezone and GPS got removed :(
Wonder if its better to wait for this fix or manually fix everything.

CyberTimon:
Thanks a lot for the PR! Looks good to me.

cl1x:
Did that really fix the GPS removal? I still have the issue in 1.6.2

Issue #1371, opened 2026-07-13
BUG: Malformed EXIF serialization in .rrdata layer leaks string array padding, breaking lens auto-selection


QuillDedCodez:
### What happened?

**Lens Correction Auto-selection fails with a "lens profile not found" error due to malformed metadata serialization.**

When inspecting the generated `.rrdata` file, the `LensModel` field under the `exif` object incorrectly serializes fixed-size string padding or array elements. This leaks escaped internal quotation marks, trailing commas, and empty string buffers directly into the JSON string value:

**Malformed data in `.rrdata`:**

```json
{
  "exif": {
    "LensModel": "\"EF-S18-55mm f/3.5-5.6 IS STM\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\", \"\""
  }
}

```

<img width="559" height="179" alt="Image" src="https://github.com/user-attachments/assets/d0575699-6eb6-49ce-ad11-517ee88234f4" />

### What did you expect to happen?

The `LensModel` field should be cleanly parsed and serialized as a single, sanitized string primitive without trailing array elements or escaped string markers.

**Expected format in `.rrdata`:**

```json
{
  "exif": {
    "LensModel": "EF-S18-55mm f/3.5-5.6 IS STM"
  }
}

```

Sanitizing this payload will allow the lens correction engine to successfully match the lookup signature and auto-select the profile.

---

<img width="321" height="347" alt="Image" src="https://github.com/user-attachments/assets/858a950b-8e73-4d1a-b6f3-051365cee709" />

### Operating System

Windows 11

### RapidRAW Version

v1.5.9

### Additional information

The underlying metadata extraction engine, `dnglab`, handles the source file correctly. Running `dnglab analyze` natively shows that the source metadata contains clean, typed primitives without corruption:

```bash
dnglab analyze --meta ".\...\IMG_3369.CR2"

```

**Truncated `dnglab` JSON output:**

```json
{
  "lens_serial_number": "xxx",
  "lens_make": "Canon",
  "sub_sec_time": "75",
  "sub_sec_time_original": "75",
  "sub_sec_time_digitized": "75",
[...]

Issue #1367, opened 2026-07-13
BUG: Manage your lenses in settings


bakman2:
### What happened?

Not sure if this is a bug - but this states 'Manage your lenses in settings'  - where can i find this exactly ?

<img width="323" height="150" alt="Image" src="https://github.com/user-attachments/assets/1f0f028f-6986-49bf-b87f-57091e049bb7" />

My lens is not detected automatically  (Nikkor Z DX16-50 f/3.5-6.3 VR) so i want to saved it - but i cannot seem to find any setting for it anywhere ?

### What did you expect to happen?

Some kind of save option when a lens was manually selected. 

### How can we reproduce this?

Use lens correction.

### Operating System

macOS Apple Silicon

### RapidRAW Version

1.5.8

### Acknowledgement

- [x] I have verified that my camera and lenses are supported by the external libraries listed above, or my issue is unrelated to camera/lens support.
- [x] I have searched all existing issues, both open and closed, and did not find a duplicate.

bakman2:
Ok I found it - the settings button is only available at the 'home' page. 

This is a weird location for this settings icon/menu. 

Actually I fail to understand this 'home' page - why not just jump to the main screen - you can add folders there as well and add the settings icon to the menu bar as well. Make it at least an checkbox option to 'always continue session' or something.

CyberTimon:
Thanks for the feedback! :)

Issue #1266, opened 2026-06-09
BUG: Lens autodetection doesn't work with Canon EOS M50 + Sigma 56mm F1.4


RigleGit:
### What happened?

When opening a .CR3 file taken with a Canon EOS M50 and a Sigma 56mm F1.4 DC DN | Contemporary 018, RapidRAW does not correctly auto-select the lens model.

ExifTool detects the lens model correctly:

```bash
exiftool -Make -Model -LensModel -Lens -FocalLength -FNumber -SubjectDistance _MG_0569.CR3 
Make : Canon 
Camera Model Name : Canon EOS M50 
Lens Model : 56mm F1.4 DC DN | Contemporary 018 
Lens : 56.0 mm 
Focal Length : 56.0 mm 
F Number : 1.4
```

The lens is a Sigma lens, but the camera maker is Canon, so it looks like RapidRAW may be trying to match it as a Canon lens instead of finding the Sigma profile.

### What did you expect to happen?

I expected the program to automatically match the lens to the existing Sigma profile, something like "Sigma 56mm F1.4 DC DN | Contemporary 018" or "Sigma 56 mm F1.4 DC DN | C 018 for Canon EF-M mount".

### How can we reproduce this?

1. Open a .CR3 file shot with a Canon EOS M50 and the lens.
2. Check the lens correction / lens auto-selection.
3. RapidRAW does not correctly auto-select the Sigma lens profile.

### Operating System

macOS Apple Silicon

### RapidRAW Version

v1.5.7

### Additional information

I checked the Lensfun database and the Sigma 56mm profile seems to already exist, so I think this isn't a missing lens profile issue.

I think it's related to the lens matching logic. Since the EXIF Make field is Canon, RapidRAW could be trying to match the lens only against Canon lenses. However, in this case the actual lens is Sigma, but the Lens Model field does not include the Sigma maker name, as you can see in the command I mentioned before.

Maybe the auto-detection could compare both:
* the best match for the camera maker
* the best global match across all lens makers

and then choose the global match if it is clearly better (or if there's no clear option within the camera maker).

Thanks for this great software! By far the best alternative to Lightroom I've found 👍🏻 

### Acknowledgement

- [x] I have verified that my camera and lenses are supported by the external libraries listed above, or my issue is unrelated to camera/lens support.
- [x] I have searched all existing issues, both open and closed, and did not find a duplicate.

XDA-Bam:
[...]

Issue #1210, opened 2026-05-27
FEATURE: Auto apply lens correction option


micahlt:
### Feature Description

Re: https://github.com/CyberTimon/RapidRAW/issues/626 - there should be an option in global settings to automatically apply lens corrections when a supported lens was used.  This will dramatically speed up the workflow for editing, as it's somewhat tedious to manually correct all of the photos in a large batch.

### Additional Context

If possible, there should be a feature for providing defaults for distortion, chromatic aberration, and vignetting settings for lens correcting.  I have a lens that forces me to turn the distortion correction slider up to 200 in order for the image to come out correctly.

### Similar feature requests

- [x] I have searched existing issues and discussions and didn't find a duplicate

Duecki1:
Like this?

When a know lens is used, the lens profile is applied automatically:

https://github.com/user-attachments/assets/7fbc700a-a6f4-4ba7-b7b7-a7acc4314bce

And when the Lens is not know, it asks the user to manually apply a profile:

https://github.com/user-attachments/assets/62763d89-3984-443f-818f-f9eb85f90d26

micahlt:
I think maybe asking the user to choose a lens if there isn't a supported one might be a little too aggressive.  I have a lens that isn't supported, and if I was asked to pick a lens to correct every single time I open a photo, that would be super annoying.  Maybe just don't try to apply corrections at all if the photo was taken on an unsupported lens?

Duecki1:
> I think maybe asking the user to choose a lens if there isn't a supported one might be a little too aggressive.  I have a lens that isn't supported, and if I was asked to pick a lens to correct every single time I open a photo, that would be super annoying.  Maybe just don't try to apply corrections at all if the photo was taken on an unsupported lens?

Hmm, I have a unsupported lens too, I would want to get that pop up to remind me to choose a lens profile. Since without it, the image looks distorted. I don't really see people who want the automatic lens correction forced, not wanting to manually apply lens corrections for an unsupported lens.

Maybe another checkbox in the settings could change that behaviour.

Or the user could have a checkbox when applying a lens correction manually "Remember these settings for this lens" or "Don't ask for this lens".

micahlt:
Those sound like helpful compromises.  I think I'd be happy with either of those approaches!

Duecki1:
> Those sound like helpful compromises. I think I'd be happy with either of those approaches!

[...]

PR #1377, opened 2026-07-14
Fix: Improve LensModel extraction and EXIF metadata serialization

QuillDedCodez:
## Description
This PR improves EXIF metadata extraction by normalizing EXIF value
representations before serializing metadata into RRData sidecar files.

The primary objective is to correctly preserve `LensModel` metadata across
different EXIF encodings, allowing RapidRAW's automatic lens correction to
reliably identify compatible lens profiles.

While implementing this fix, additional EXIF parsing paths were improved to
produce cleaner, more consistent metadata by properly handling `Ascii` and
`Undefined` EXIF values.

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made
- Added `decode_cfa_pattern()` to decode TIFF-encoded
  `CFAPattern` (`Undefined`) values into standard Bayer pattern
  strings (e.g. RGGB, BGGR).

- Added `clean_ascii_field()` to normalize EXIF `Ascii` values by removing
  null padding and empty components before writing RRData sidecar metadata.

- Added `clean_undefined_field()` to normalize EXIF `Undefined`
  values according to the EXIF specification by decoding
  `UserComment` while suppressing meaningless binary payloads
  (e.g. empty `MakerNote` data).

- Improved `LensModel` extraction by correctly handling multiple EXIF value
  representations before serializing metadata into RRData sidecar files,
  allowing automatic lens correction to detect compatible lens profiles.

- Added fallback handling for `LensSpecification` using RAW metadata when
  invalid or incomplete aperture rational values are returned by
  `kamadak-exif`.

- Improved metadata extraction consistency by sharing common parsing logic
  between standard EXIF extraction and RAW metadata extraction paths.

- Improved EXIF metadata sanitization to prevent null-padded strings,
  binary payloads, and unsupported `Undefined` values from being exported
  into RRData sidecar metadata.

### NEF RAW Image

<table align="center">
<tr>
    <td align="center"><b>Before</b></td>
    <td align="center"><b>After</b></td>
  </tr>
  <tr>
<td>Image
<td>Image
</tr>
</table>

---
<div align="center">
  <p><b>Lens Auto Correction Fix</b></p>
Image
</div>

---
### CR2 RAW Image
<table align="center">
<tr>
    <td align="center"><b>Before</b></td>
    <td align="center"><b>After</b></td>
  </tr>
  <tr>
<td>{347D685E-F133-4E5C-983A-6464539012D8}
<td>{6CA145D7-C535-415B-BDAD-0B1420B69EA1}
</tr>
</table>

---

<table align="center">
<tr>
    <td align="center"><b>Before</b></td>
    <td align="center"><b>After</b></td>
  </tr>
  <tr>
<td>{B59B8F03-0A2C-484E-B965-9383A6912B5D}
<td>{9313AFD2-F5B7-4447-9A23-6615448647BB}
</tr>
</table>

---
### JPEG Non-Raw Image

<div align="center">
  <p><b>Non-Raw Image Test</b></p>
{A9668E09-DA18-4A6E-91B4-6C2F3DA7BC90}
</div>

---

<div align="center">
  <p><b>Lens Auto Correction Fix</b></p>
{7D68B584-9C24-4C8B-9E8B-992692858CF6}
</div>

---

## Testing
**Verified:**
- [X] LensModel is correctly written to the RRData sidecar.
- [X] Automatic lens profile selection now succeeds for the sample image from #1123.
- [X] CFAPattern is decoded correctly.
- [X] ASCII EXIF fields no longer contain null-padded strings.
- [X] UserComment is decoded correctly.
- [X] Existing metadata extraction behavior remains unchanged for tested images.

**Test Configuration:**
- **OS:** Windows 11 24H2 LTSC
- **Hardware:** RYZEN 5600G, RTX 4060TI
## Checklist

## Additional Notes

This PR originated from investigating the EXIF metadata serialization issue
documented in #1371.

The investigation identified that EXIF metadata written to RRData sidecar files
was not consistently normalized across different EXIF value representations.
Consequently, metadata such as `LensModel` could be omitted or incorrectly
serialized, preventing RapidRAW's automatic lens correction from reliably
identifying compatible lens profiles.

To validate the fix, I tested the sample image provided in #1123. After these
changes, the generated RRData sidecar contains the expected metadata, allowing
automatic lens profile selection to function correctly for that sample.

Since the root cause affects EXIF metadata serialization, these changes may
also resolve or improve related issues involving lens detection and automatic
lens correction, including:

- #1123
- #1367
- #1210
- #1266

## AI Disclaimer:

### DIFF
--- a/src-tauri/src/exif_processing.rs
+++ b/src-tauri/src/exif_processing.rs
@@ -295,9 +295,6 @@ fn clean_undefined_field(
     fallback_display: impl FnOnce() -> String,
 ) -> Option<String> {
     if tag_name == "UserComment" {
-        // First 8 bytes are a character-code header (e.g. b"ASCII\0\0\0",
-        // b"UNICODE\0"). Strip it, then check whether what remains is
-        // meaningful text or just padding/zeros.
         let body = if bytes.len() > 8 { &bytes[8..] } else { &[][..] };
         let text = String::from_utf8_lossy(body)
             .trim_matches(char::from(0))
--- a/src-tauri/src/exif_processing.rs
+++ b/src-tauri/src/exif_processing.rs
@@ -132,6 +132,29 @@ fn parse_raw_creation_date(date_str: Option<&str>) -> Option<DateTime<Utc>> {
     parse_creation_datetime(date_str?).map(|dt| DateTime::from_naive_utc_and_offset(dt, Utc))
 }
 
+fn clean_ascii_value(value: &exif::Value) -> Option<String> {
+    let exif::Value::Ascii(ref components) = *value else {
+        return None;
+    };
+
+    let cleaned: Vec<String> = components
+        .iter()
+        .map(|c| {
+            String::from_utf8_lossy(c)
+                .trim_matches(char::from(0))
+                .trim()
+                .to_string()
+        })
+        .filter(|s| !s.is_empty())
+        .collect();
+
+    if cleaned.is_empty() {
+        None
+    } else {
+        Some(cleaned.join(" "))
+    }
+}
+
 pub fn read_exif(file_bytes: &[u8]) -> Option<Exif> {
     let exifreader = exif::Reader::new();
     exifreader
@@ -332,12 +355,19 @@ pub fn extract_metadata(file_bytes: &[u8]) -> Option<HashMap<String, String>> {
                         fmt_date_str(field.display_value().to_string()),
                     );
                 }
-                _ => {
-                    let val = field.display_value().with_unit(&exif_obj).to_string();
-                    if !val.trim().is_empty() {
-                        map.insert(field.tag.to_string(), val);
+                _ => match &field.value {
+                    exif::Value::Ascii(_) => {
+                        if let Some(val) = clean_ascii_value(&field.value) {
+                            map.insert(field.tag.to_string(), val);
+                        }
                     }
-                }
+                    _ => {
+                        let val = field.display_value().with_unit(&exif_obj).to_string();
+                        if !val.trim().is_empty() {
+                            map.insert(field.tag.to_string(), val);
+                        }
+                    }
+                },
             }
         }
     }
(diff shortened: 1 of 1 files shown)


CyberTimon:
Hey @QuillDedCodez, thanks for putting this together!

I can't merge this just yet though, as a few of these changes introduce some problems.

Here is what needs to be fixed (from a quick test & review):

* In `extract_metadata`, if the basic EXIF is empty but your new lens patch succeeds, the early map empty check will return immediately. This completely skips extracting Make, Model, and other critical fallback data for RAWs.
* In `clean_undefined_field`, you're checking if the byte array is all zeros *before* checking if the tag name is `MakerNote`. `MakerNote` tags are often several megabytes. Because Rust evaluates eagerly, this forces a massive byte-by-byte scan on every image load. Put the tag name check first so it early exits!
* You're calling `read_raw_metadata` early just to look for the lens spec. This fires up the heavy RAW parser for every single JPEG/PNG that lacks a lens tag. Please move this down into the existing RAW fallback block so we don't parse files twice.
* `clean_ascii_field` uses `find()`, which drops everything after the first string. If a field like copyright or description is an array of strings, the rest get lost. Let's filter and join them instead like the old code did.

A few minor things to tweak while you're at it:
* `UserComment` shouldn't blindly assume UTF-8. The first 8 bytes dictate the encoding (ASCII, UNICODE, etc). Assuming UTF-8 will heavily mangle UTF-16 comments.
* `decode_cfa_pattern` hardcodes little-endian byte reading, which will fail on big-endian (Motorola) TIFFs.
* Comparing focal length floats with `f32::EPSILON` won't work well for magnitudes > 1.0 due to how floating-point math works. Use a flat tolerance (like `0.01`) or just check exact equality.

Thanks so much & have a nice evening,
Timon

QuillDedCodez:
Hey Timon, thanks for the review — reworked this rather than patching the original diff.

Scope now: `LensModel`/`Make`/`Model` corruption and `LensSpecification` formatting. Dropping `MakerNote`, `UserComment`, and `CFAPattern` — not confident enough in those yet, and didn't want to bundle unverified logic into a fix for a specific, confirmed bug. `MakerNote` is proprietary per manufacturer anyway, so out of scope regardless. Can follow up separately once I've validated them properly.

- **`is_empty()` gate**: Fixed — lens-spec logic never inserts before that check.
- **`find()` dropping data**: Fixed — filters and joins every component now, not just the first.
- **`f32::EPSILON`**: Fixed — flat `0.01` tolerance for both focal length and aperture.
- **Double parsing**: Added a second `read_raw_metadata()` call, gated to only fire when focal parsed but aperture didn't — mutually exclusive with the existing fallback, so never double-fires per file. It's a full metadata re-parse when it does fire, not a targeted lookup (`rawler` has no field-level extraction), but it never touches pixel data, so it's not the "heavy RAW parser" cost from your original comment.

  Added it after finding a CR2 where kamadak-exif reads this tag's aperture as `0/0` (→ `NaN` in output) while rawler's parse of the same file is valid, confirmed via `dnglab`. Root cause of the discrepancy is still open.

Let me know if that tradeoff's acceptable, or if you'd rather avoid the full re-parse even in this narrow case.

CyberTimon:
LGTM! Thanks a lot for the update :)

PR #347, opened 2025-09-24
add memory-mapped file loading and prepare exclusive file locking

riccio8:
This change sets up the groundwork for opening and reading files

- Load files into memory using memmap2
- Introduce future exclusive locking (currently working on this)
- Foundation to prevent files from being modified during thumbnail generation

related: https://github.com/CyberTimon/RapidRAW/issues/307 - https://github.com/CyberTimon/RapidRAW/issues/330

PR #1387, opened 2026-07-16
fix: prevent SIGBUS crash when reading images from external volumes

olegsatov:
## Summary

RapidRAW crashes with `SIGBUS` (`EXC_BAD_ACCESS` / `KERN_MEMORY_ERROR`) when an image file is memory-mapped and the kernel later fails to page it in — e.g. an external/USB drive disconnects or returns an I/O error, a network share drops, or the file is truncated/replaced by another process after mapping.

Crash report (macOS 26.4, ARM64), faulting thread:

```
_platform_memmove
rawler::rawsource::RawSource::new_from_slice
rapidraw_lib::raw_processing::develop_raw_image
rapidraw_lib::image_loader::load_base_image_from_bytes
```

Kernel triage: `VM - Page has error bit set`, and the faulting address lies **inside** a valid `mapped file` region — i.e. not a bad pointer, but a failed page-in from the backing file. In my case two worker threads faulted simultaneously on two different mapped files when the external drive the library lives on glitched.

Since this is a signal, not a panic, `catch_unwind` can't contain it and the whole app dies. The existing "fall back to `fs::read`" branches only trigger when creating the mapping fails — they can't help once a page-in fails later.

## Fix

Replace `read_file_mapped` with `read_file_bytes` — a plain `fs::read` with the same pre-checks — at all call sites (image load, thumbnails, EXIF batch read, preview, export, negative conversion). I/O errors now surface as regular `Err` results that the UI already handles, instead of killing the process.

No performance regression is expected: every consumer (`rawler::RawSource::new_from_slice`, image decoders) immediately copies the entire buffer into an owned allocation anyway, so mmap never saved a copy here — it only added the SIGBUS risk.

Also removes the now-unused `memmap2` dependency and the `ReadFileError::Locked` variant (the advisory `try_lock_shared` never meaningfully protected these reads).

## Testing

- `cargo check` clean, no new warnings (on top of current `main`).
- I could not artificially reproduce the original crash — it requires a storage-level fault (drive disconnect / I/O error) mid-load. After this change the failure mode is a normal read error returned to the UI instead of a process crash.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -34,7 +34,6 @@ dependencies = [
  "libc",
  "little_exif",
  "log",
- "memmap2",
  "mimalloc",
  "mozjpeg-rs",
  "nalgebra",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -51,7 +51,6 @@ tempfile = "3.27.0"
 mimalloc = "0.1.52"
 image_hasher = "3.1.1"
 regex = "1.12.4"
-memmap2 = "0.9.11"
 half = { version = "2.7.1", features = ["bytemuck"] }
 glam = "0.33.2"
 quick-xml = { version = "0.41", features = ["serialize"] }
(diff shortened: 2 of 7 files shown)


CyberTimon:
What do you think about performance? Will this have a huge effect on NAS drives?

For context see https://github.com/CyberTimon/RapidRAW/pull/347

Issue #1342, opened 2026-07-07
BUG: Exporting JPEG XL causes massive memory consumption


distinctjuggle:
### What happened?

Exporting files as JPEG XL with quality 100 (lossless) causes >40 GiB of memory to be consumed for 15 images.  Memory usage seems to increase until the kernel kills RapidRaw.

### What did you expect to happen?

Proper export

### Operating System

Other Linux

### RapidRAW Version

1.5.8

### Additional information

Arch Linux AUR package `rapidraw-bin 1.5.8-1`

### Acknowledgement

- [x] I have verified that my camera and lenses are supported by the external libraries listed above, or my issue is unrelated to camera/lens support.
- [x] I have searched all existing issues, both open and closed, and did not find a duplicate.

AkalaAkala:
Still present

Redo11:
For a single big file, the memory usage can exceed 50GB for me, however it does export it eventually (if not killed by OOM killer)

PR #1395, opened 2026-07-18
feat(gpu): add bounded high-precision export rendering

VailElla:
## Description

Add a bounded, crate-private high-precision GPU render path for the follow-up JPEG XL export work. The normal preview and export path keeps its existing 8-bit contract, while the new path renders tile output as `rgba16float` and returns RGB/RGBA 32-bit float image data without an 8-bit intermediate.

This draft also bounds reusable compute textures to one overlapped processing tile instead of allocating six full-image textures for large inputs.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Performance improvement
- [x] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made

- Added a high-precision compute pipeline variant with an `rgba16float` storage target while deferring integer dithering to the follow-up encoder.
- Generalized tiled GPU readback for both 4-byte RGBA8 and 8-byte RGBA16F pixels.
- Converted RGBA16F readback into RGB32F or RGBA32F while preserving the source alpha contract.
- Kept the high-precision entry crate-private until the follow-up JXL backend wires it into export.
- Bounded reusable ping-pong, blur, and tile-output textures to a `2304 × 2304` overlapped tile; full-size display textures remain unchanged.
- Avoided a full RGBA32F intermediate when the input is already RGB32F or RGBA32F.
- Added three unit tests for high-precision shader behavior, RGB32F upload conversion, and the bounded tile extent.

## Screenshots/Videos

N/A — this is a backend-only foundation and does not change the current UI or exported output by itself.

## Testing

**Test Configuration:**

- **OS:** macOS 26.5.2
- **Hardware:** Apple M4 Pro, 14 logical cores, 48 GiB RAM

**Automated:**

- `cargo +1.96.1 fmt --manifest-path src-tauri/Cargo.toml --check` — passed
- `cargo +1.96.1 check --manifest-path src-tauri/Cargo.toml` — passed; warning set matches current `main` with no new warnings
- `cargo +1.96.1 test --manifest-path src-tauri/Cargo.toml -- --nocapture` — 3 passed, 0 failed
- `cargo +1.96.1 test --manifest-path src-tauri/Cargo.toml gpu_processing:: -- --nocapture` — 3 passed, 0 failed after final cleanup
- `cargo +1.96.1 clippy --manifest-path src-tauri/Cargo.toml --all-targets --all-features` — passed; only current macOS baseline warnings remain
- `npm ci` — completed with 0 vulnerabilities
- `npm run build` — passed
- `git diff --check` — passed

## Checklist

## Additional Notes

- Based directly on `main` at `41d92c8f` and kept to two focused commits touching only `src-tauri/src/gpu_processing.rs`.
- This is an internal prerequisite for the smaller JXL export series related to #1322 and #1342; it does not close either issue by itself.
- The crate-private high-precision entry is intentionally not wired into export in this draft. JXL encoding, metadata, export controls, and end-to-end real-image validation remain in follow-up PRs.
- The existing 8-bit preview/display path remains the default and retains its current output format.

## AI Disclaimer:

### DIFF
--- a/src-tauri/src/gpu_processing.rs
+++ b/src-tauri/src/gpu_processing.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use half::f16;
-use image::{DynamicImage, GenericImageView, ImageBuffer, Luma, Rgba};
+use image::{DynamicImage, GenericImageView, ImageBuffer, Luma, Rgb, Rgba};
 use std::num::NonZero;
 
 #[cfg(not(any(target_os = "android", target_os = "linux")))]
@@ -12,6 +12,18 @@ use wgpu::util::{DeviceExt, TextureDataOrder};
 use crate::image_processing::{AllAdjustments, GpuContext, MAX_MASKS};
 use crate::lut_processing::Lut;
 use crate::{AppState, GpuImageCache};
+
+fn high_precision_shader_source() -> String {
+    include_str!("shaders/shader.wgsl")
+        .replace(
+            "texture_storage_2d<rgba8unorm, write>",
+            "texture_storage_2d<rgba16float, write>",
+        )
+        .replace(
+            "let dither_amount = 1.0 / 255.0;",
+            "let dither_amount = 1.0 / 65535.0;",
+        )
+}
 
 #[derive(Clone, Copy, Debug)]
 pub struct Roi {
@@ -418,8 +430,9 @@ fn read_texture_data_roi(
     texture: &wgpu::Texture,
     origin: wgpu::Origin3d,
     size: wgpu::Extent3d,
+    bytes_per_pixel: u32,
 ) -> Result<Vec<u8>, String> {
-    let unpadded_bytes_per_row = 4 * size.width;
+    let unpadded_bytes_per_row = bytes_per_pixel * size.width;
     let align = wgpu::COPY_BYTES_PER_ROW_ALIGNMENT;
     let padded_bytes_per_row = (unpadded_bytes_per_row + align - 1) & !(align - 1);
     let output_buffer_size = (padded_bytes_per_row * size.height) as u64;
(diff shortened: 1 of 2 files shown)


VailElla:
Refresh validation (2026-07-25): rebased onto current upstream main and force-with-lease updated. Added an adapter capability gate for the RGBA16Float high-precision path.

Compatibility behavior:
- adapters whose RGBA16Float format supports texture binding, storage binding, and copy-source create the high-precision pipeline;
- other adapters keep the existing RGBA8 preview/export pipeline, while a request for the internal high-precision path returns a specific unsupported-GPU error instead of preventing normal GPU initialization.

Memory regression coverage now models a 100 MP (11,600 x 8,600) input: five reusable RGBA16F intermediates remain capped at 2,304 x 2,304 (<220 MiB combined), versus >3 GiB if those intermediates were full-image allocations. Full-size RGBA8 display textures are intentionally unchanged.

`cargo test gpu_processing:: --lib` passed: 4 passed, 0 failed.

akhater:
We shipped a high-precision export path this week and spent a while with this PR while deciding what to build, so here is what we found — some of it is probably useful to you, and one part of it may have been overtaken.

**The bounded textures look like they already landed.** Timon's `62ea75bf` ("clamp intermediate texture sizes to tile bounds to fix VRAM OOM on export") clamps the ping-pong, blur and tile-output textures to the overlapped tile, which is the same fix as the second half of this PR. Worth a rebase to see how much of yours is left standing — it would make the remaining diff a lot easier to review.

**On `rgba16float` as the target.** For the JXL work this is fine, since JXL carries float. For an integer container it is not quite enough: half-float has 11 significant bits, so the brightest stop of an encoded signal lands on a grid of roughly 32 steps out of 65535. Invisible on screen, and not what a 16-bit file promises. We went with `rgba32float` for that reason. The tile texture is the only one that has to change format — `working_texture` and `output_texture` are written solely behind `output_to_display`, so an export can leave them at `Rgba8Unorm`, or at 1×1, and the extra VRAM is one tile rather than several full frames.

**Three things that bit us and are easy to miss when this gets wired into an encoder:**

1. The dither in `main()` adds ±0.5/255 before the store. On a high-precision target that is 257× too large and has to be gated off, which @dimafa does neatly with a pipeline constant in #1466.
2. `apply_watermark` composites with `image::imageops::overlay`, which goes through `GenericImage for DynamicImage` — pixel type `Rgba<u8>`. On a float image it quantises **every pixel in the stamp's bounding box, including the fully transparent ones**, so a watermark at zero opacity silently demotes a rectangle of the photograph.
3. `export_masks_for_image` renders on its own call, so per-mask files keep the old 8-bit path unless they are routed too. We missed that on the first pass and shipped one honest file with N dishonest companions beside it in a test batch.

**One more, for whoever tests this.** Our first real GPU test rendered pure black and reported it as a precision failure. `AllAdjustments::default()` zeroes the camera-profile matrix, and the shader multiplies by it — so the fixture has to build adjustments through the real JSON path or it proves nothing. Worth knowing before you chase it in the pipeline.

To be straight about it: we did not take code from this PR, mostly because of the overlap with `62ea75bf` above and because we needed a 32-bit target. The idea we did build on is @dimafa's in #1466 — rewriting the storage format in the shader source and gating the dither behind a pipeline constant — and that PR is the one we would like to see land upstream. Ours is in [`mods/export_precision.rs`](https://github.com/akhater/Argentum/blob/main/src-tauri/src/mods/export_precision.rs) if any of it is useful to lift, licence and all; it is in Argentum, our prime sharp RapidRAW fork.

PR #1411, opened 2026-07-23
Improve Library Browsing Performance

### LINKED ISSUE #1391: BUG: abysmal startup performance on "spinning rust" HDD storage
### What happened?

Whenever i open a new directory or resume session with RapidRaw, i have to spend ~5-10 minutes waiting with high I/O until rapidraw starts responding and producing thumbnails.

Sometimes the main window won't open until directory is scanned first.

It would be nice to either set lower amount of background processing workers or simply delegate file processing to be on-demand, for people who are more economical with their price-per-gigabyte storage choices.

a more performant behaviour when displaying directory listing

1. prepare a directory of ~700 CR3 / dng files on HDD  (i have cr3 files from canon r7 and i tried converting them with dnglab as well)
2. open said directory in RapidRAW
3. express saintly patience
4. rapidraw becomes reponsive again, eventually

### Operating System

Other Linux

### RapidRAW Version

1.5.9 

### Additional information

I am using prebuilt deb release on Gentoo Linux

### Acknowledgement

RayJW:
## Description
This started as an approach to fix #1391, and then became something a little bigger.
I made a few changes that should greatly improve the library performance, the gains should be the biggest on HDDs, but also apply to SSDs.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Performance improvement
- [X] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made
- Limit the worker threads for thumbnail loading on spinning disks to just one process
- Load thumbnails in path order, to make better use of sequential performance and lose less to random access
- Load the embedded preview for RAW files if possible to avoid parsing the full RAW unless necessary

## Testing

**Test Configuration:**
- **OS:** CachyOS
- **Hardware:** Ryzen 9, 64 GB RAM, AMD GPU

## Checklist

## Additional Notes
The change about embedded previews kinda changes the RapidRAW behaviour. Because we use the thumbnail generated by the camera, there is some processing etc. applied.
That means when actually switching to the image it will look different, because we actually load the RAW on that action.
Personally I think it is more than worth the trade-off for the improved performance and decreased compute effort, however, it could be argued that we should make this a preference, in case people dislike that behaviour.
Please let me know if there are any strong opinions on that, and I can add a preference.

## AI Disclaimer:
- [X] This PR is AI-generated but guided by a human

### DIFF
--- a/src-tauri/src/app_state.rs
+++ b/src-tauri/src/app_state.rs
@@ -6,6 +6,7 @@ use std::sync::{Arc, Condvar, Mutex};
 
 use image::{DynamicImage, GrayImage};
 use serde::{Deserialize, Serialize};
+use sysinfo::Disks;
 use tokio::sync::Mutex as TokioMutex;
 use tokio::task::JoinHandle;
 use wgpu::{Texture, TextureView};
@@ -174,4 +175,6 @@ pub struct AppState {
     pub decoded_image_cache: Mutex<DecodedImageCache>,
     pub thumbnail_manager: Arc<ThumbnailManager>,
     pub metadata_manager: Arc<MetadataManager>,
+    pub disks_cache: Mutex<Option<Disks>>,
+    pub disks_cache_refreshing: AtomicBool,
 }
--- a/src-tauri/src/app_state.rs
+++ b/src-tauri/src/app_state.rs
@@ -101,6 +101,8 @@ pub struct ThumbnailManager {
     pub queue: Mutex<VecDeque<String>>,
     pub cvar: Condvar,
     pub processing_now: Mutex<HashSet<String>>,
+    pub rotational_disk: AtomicBool,
+    pub io_gate: Mutex<()>,
 }
 
 impl ThumbnailManager {
@@ -109,6 +111,8 @@ impl ThumbnailManager {
             queue: Mutex::new(VecDeque::new()),
             cvar: Condvar::new(),
             processing_now: Mutex::new(HashSet::new()),
+            rotational_disk: AtomicBool::new(false),
+            io_gate: Mutex::new(()),
         })
     }
 }
--- a/src-tauri/src/file_management.rs
+++ b/src-tauri/src/file_management.rs
@@ -390,12 +390,7 @@ pub async fn update_exif_fields(
     .map_err(|e| format!("Task failed: {}", e))?
 }
 
-fn is_rotational_disk(path: &Path) -> bool {
-    let Ok(canonical) = path.canonicalize() else {
-        return false;
-    };
-
-    let disks = Disks::new_with_refreshed_list();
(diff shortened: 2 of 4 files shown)


CyberTimon:
Hey @RayJW 

Are you aware that calling `Disks::new_with_refreshed_list()` inside sync Tauri commands runs on the main thread and can freeze the UI if a drive is slow/sleeping? Probably making the command async or caching disk types on startup would fix it??

Also, I'd like to add a toggle setting for embedded RAW JPEG extraction (enabled by default). If you can handle the blocking issue, I'm happy to add the toggle option myself!

Thanks a lot for your contribution :)

Timon

RayJW:
> Are you aware that calling `Disks::new_with_refreshed_list()` inside sync Tauri commands runs on the main thread and can freeze the UI if a drive is slow/sleeping? Probably making the command async or caching disk types on startup would fix it??

Good call, I've rewritten the functionality to poll on startup or if there is a new disk detected and otherwise only pull from the cached results. Feel free to add the settings toggle or let me know if you want me to do it :)

CyberTimon:
Thanks for fixing the blocking issue, the non-blocking cache looks great! Since _io_permit holds the lock through the entire generate_single_thumbnail_and_cache call, it locks the CPU image decoding/resizing alongside the disk read. If we scope the lock strictly around the file read, disk reads stay strictly sequential (no thrashing), but multi-core CPUs can decode images in parallel from RAM. It feels a bit over-throttled, what do you think?

RayJW:
> Thanks for fixing the blocking issue, the non-blocking cache looks great! Since _io_permit holds the lock through the entire generate_single_thumbnail_and_cache call, it locks the CPU image decoding/resizing alongside the disk read. If we scope the lock strictly around the file read, disk reads stay strictly sequential (no thrashing), but multi-core CPUs can decode images in parallel from RAM. It feels a bit over-throttled, what do you think?

Good catch! I changed the code to warm the OS page cache with a sequential read while holding the gate.

CPU decode runs should now be parallel again, however it relies on the page staying cached. If under heavy memory pressure the page gets evicted before the decode stage reads it, we'll hit an unguarded disk read. Shouldn't be an issue in practice though, since decode happens immediately after prefetch on the same thread, well within typical page cache lifetimes.

RayJW:
I've tacked on one very minor thing that is also related to the performance of the library view. By using the callback we can avoid re-creating the `getItemSize` function every time the view is rendered. I don't expect this to have a measurable impact to be honest, but I still think it's the correct thing to do and even miniscule gains still add up sooner or later :)

If this feels to out of scope I'll happily revert it :)

CyberTimon:
Amazing, looks good to me now :) Thanks for all the work.

PR #1413, opened 2026-07-23
Add Ctrl/Cmd+F keybind to focus the active search field

magnusrodseth:
## Description

Pressing `Ctrl+F` (`Cmd+F` on macOS) now focuses the search field of the view you are in: the library search in the main library, or the preset search on the Community page. The action is called `focus_search`, and since it goes through the existing keybind registry it shows up in Settings > Keybinds and is rebindable like every other action. In the editor the shortcut currently does nothing (there is no obvious search target there yet, see Additional Notes).

**Why it is built this way:**

- **Registry first.** The action is a normal `KEYBIND_DEFINITIONS` entry, so the settings UI, user overrides, and the mac/win `ctrl`/`cmd` unification all work without any special-casing. `Ctrl+F` was unbound, so no existing binding changes.
- **A focus-request counter instead of refs across the tree.** The two candidate search inputs own their state locally and are never mounted at the same time (`MainLibrary` and `CommunityPage` are mutually exclusive on `activeView`). So the shortcut handler just increments a `searchFocusRequest` counter in `useUIStore`, and whichever search input is currently mounted reacts to the change and focuses itself. No new context, no ref plumbing through `App.tsx`.
- **Mount guard.** Each input compares against the counter value it saw on mount, so navigating back to the library does not replay an old request and steal focus.
- **No focus-stealing while typing.** The existing input-focus guard in `useKeyboardShortcuts` already ignores shortcuts while an input is focused, so `Ctrl+F` inside a text field stays inert.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [x] UI/UX improvement

## Changes Made
- `src/utils/keyboardUtils.ts`: new `focus_search` definition, default combo `Ctrl/Cmd+F`, section `library`
- `src/store/useUIStore.ts`: `searchFocusRequest` counter and `requestSearchFocus()` action
- `src/hooks/useKeyboardShortcuts.ts`: `focus_search` handler; fires only outside the editor (`!selectedImage`)
- `src/components/panel/library/LibraryHeader.tsx`: `SearchInput` expands and focuses on request
- `src/components/panel/CommunityPage.tsx`: preset search input focuses on request
- `src/i18n/locales/*.json`: `settings.keybinds.actions.focus_search` label added to all 12 locales

## Screenshots/Videos

N/A (the visible behavior is the existing search field expanding and receiving focus).

## Testing

Verified on a local dev build (`npm start`) by driving the app with scripted keystrokes plus temporary console instrumentation (removed before commit):

- In the library, `Cmd+F` dispatches `focus_search`, the search field expands, and `document.activeElement` is confirmed to be the search input; typed text lands in the field.
- While the search input is focused, a second `Cmd+F` never reaches the shortcut handler (the existing input-focus guard swallows it), so no re-trigger loops.
- In the editor, `Cmd+F` maps to `focus_search` but does not fire (`shouldFire` requires no selected image).
- Mounting `SearchInput` with a stale request counter does not steal focus (the mount guard works).

The Community page input uses the identical mechanism but was exercised only through the library path. Static checks: `tsc --noEmit`, `eslint`, and `npm run i18n:check` compared against `main` show no new findings introduced (see note below); `npm run build` and `prettier --check` pass.

**Test Configuration:**
- **OS:** macOS 26.5.2
- **Hardware:** Apple M3 Max

## Checklist

## Additional Notes

- The `focus_search` handler uses the same `(e: any, s: any)` signature as every other handler in `useKeyboardShortcuts.ts`, which means eslint counts 3 more `no-explicit-any` hits in a file that already has 115. I kept it consistent with the file rather than typing one handler differently; happy to adjust if you prefer.
- Follow-up ideas kept out of scope on purpose: focusing panel-level filters in the editor (for example the presets panel), and a `Ctrl/Cmd+P` command palette. I have opened a separate feature request for the palette to check appetite before building it.

## AI Disclaimer:


CyberTimon:
Thanks, looks good to me!

PR #1424, opened 2026-07-25
Add Cmd/Ctrl+L shortcut to copy image paths to the clipboard

magnusrodseth:
## Description

Adds a `Cmd+L` / `Ctrl+L` shortcut that copies the absolute path of the active image to the clipboard. Useful for pasting a file into a terminal, a chat message, or another application without hunting for it in Finder/Explorer first.

The shortcut works both while editing an image and from the library grid.

## Type of Change
- [x] New feature

## Changes Made
- Added a `copy_image_path` keybind to `KEYBIND_DEFINITIONS`, defaulting to `ctrl` + `KeyL`. Because `normalizeCombo` already folds `metaKey`/`ctrlKey` into `ctrl`, this is `Cmd+L` on macOS and `Ctrl+L` elsewhere. It picks up the existing Settings UI and rebinding support for free, and it does not collide with any other default combo.
- Implemented the action in `useKeyboardShortcuts`. The target resolves to the open image when editing, otherwise to the library multi-selection, otherwise to the active thumbnail.
- Multi-selections copy one path per line, ordered to match the visible grid rather than click order.
- Virtual copy suffixes (`?vc=`) are stripped so the clipboard receives a real filesystem path. This mirrors what `show_in_finder` already does via `parse_virtual_path` on the Rust side. Duplicates that collapse to the same physical file are removed.
- Copying uses `navigator.clipboard.writeText`, the same API `MetadataPanel` already uses for copying metadata values. No new dependency, and no new Tauri command or capability.
- Added `settings.keybinds.actions.copy_image_path` and `library.toasts.{copiedPath,copiedPaths,failedCopyPath}` to all 12 locales.

The existing `isInputFocused` guard means the shortcut does not fire while typing in the search field.

## Screenshots/Videos

No screenshot included, since the visible result is a standard toast in the existing bottom-right container.

## Testing

Verified manually: single image in the editor, single image in the library grid, a multi-selection copying newline-separated paths in grid order, a virtual copy resolving to its physical path, and no firing while the search field has focus.

Also checked against the repo's own tooling:
- `npm run build` succeeds.
- `npm run typecheck` reports the same 67 pre-existing errors before and after this change, none in the touched files.
- `npm run i18n:check` reports the same 37 pre-existing missing keys before and after. None of the keys added here are among them.
- `npm run i18n:runtime-check` passes.
- Prettier clean.

**Test Configuration:**
- **OS:** macOS 26.5.2
- **Hardware:** Apple M3 Max

## Checklist

## Additional Notes

One deliberate choice worth a reviewer's attention: the success feedback uses `toast.success`, which is the first non-error toast in the codebase. Every one of the 52 existing calls is `toast.error`. It renders through the same `ToastContainer` and the same `toastClassName`, and since that class function ignores the toast type, the background, border, and text styling are identical to existing toasts. Happy to switch it to a neutral `toast()` or drop the success message if you'd prefer to keep toasts error-only.

The two toast strings avoid an i18next plural key, using separate `copiedPath` and `copiedPaths` messages without a count. This keeps the phrasing grammatical across all 12 locales without needing per-language CLDR plural forms for a transient confirmation message.

## AI Disclaimer:


CyberTimon:
Thanks for the PR!

Please remove all toast i18n strings, except for the settings keybinds entry and remove the success toast (you can keep the english-only/untranslated error toast). Thanks a lot!

magnusrodseth:
Thanks for the review! All three done in 08bc475d:

- **Removed all toast i18n strings** — the `library.toasts` block (`copiedPath`, `copiedPaths`, `failedCopyPath`) is gone from all 12 locales.
- **Kept the settings keybinds entry** — `settings.keybinds.actions.copy_image_path` stays translated everywhere, so the binding is still listed and rebindable.
- **Removed the success toast**, and the error toast is now untranslated English (``toast.error(`Failed to copy path: ${err}`)``), matching how the rest of the codebase reports errors.

`useTranslation` is no longer needed in `useKeyboardShortcuts.ts`, so that import came out too. `npm run i18n:runtime-check` passes (952 plural resolutions across 12 locales), and typecheck/lint counts are unchanged from before the commit.

CyberTimon:
Thanks for the update!

PR #1426, opened 2026-07-26
Support importing multiple preset files at once

jhorakcz:
## Description

Preset import was limited to one file per dialog. This allows selecting several at once, mixing `.rrpreset` and legacy `.xmp`/`.lrtemplate` in the same selection, and collects a per-file result instead of aborting on the first file that fails.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made
- Added `handle_import_presets_from_files`, which reads `presets.json` once, merges every selected file, and writes once, instead of N read/write cycles.
- A file that fails to parse no longer aborts the batch. Failures are collected per file with their reason and returned alongside the updated preset list, then logged by file name.
- Extracted the parse and merge logic into shared helpers. The existing single-file commands delegate to them and keep their signatures.
- Replaced an `unwrap()` on the `.lrtemplate` regex with proper error propagation.

Per review feedback: no toasts and no translation keys. Import failures go to the console in English only, and success stays silent since the new presets appear in the panel.

**Behavior change worth flagging:** legacy `.xmp`/`.lrtemplate` import previously checked new preset names against every preset inside every folder, while `.rrpreset` import checked top-level names only. Sharing one helper meant picking one. It now checks top-level names only, matching `.rrpreset`. Presets inside a folder are namespaced by that folder, so a root-level "Vivid" and "Portraits/Vivid" are no longer treated as a collision - this removes spurious `(1)` suffixes on legacy import.

Nested sub-folders remain out of scope; `PresetFolder.children` is unchanged.

## Testing

Multi-selected four files in one dialog: a `.rrpreset` containing one root preset plus a folder of two children, a `.xmp`, a `.lrtemplate`, and a deliberately truncated `.rrpreset`. All five presets imported, the folder kept both children with working previews, and the corrupt file was logged by name with its parse error:

```
Failed to import corrupt.rrpreset: Failed to parse preset file: EOF while parsing an object at line 5 column 0
```

`cargo fmt -p RapidRAW -- --check` and `cargo clippy --all-targets --all-features -- -D warnings` both pass clean.

**Test Configuration:**
- **OS:** Windows 11

## Checklist

## Additional Notes

The two single-file commands are kept registered and working. They are now thin wrappers over the shared helpers.

## AI Disclaimer:

### DIFF
--- a/src-tauri/src/file_management.rs
+++ b/src-tauri/src/file_management.rs
@@ -243,6 +243,20 @@ pub struct PresetFile {
     pub presets: Vec<PresetItem>,
 }
 
+#[derive(Serialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct PresetImportFailure {
+    pub file_name: String,
+    pub error: String,
+}
+
+#[derive(Serialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct PresetImportResult {
+    pub presets: Vec<PresetItem>,
+    pub failures: Vec<PresetImportFailure>,
+}
+
 #[derive(Debug)]
 pub enum ReadFileError {
     Io(std::io::Error),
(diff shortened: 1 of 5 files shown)


CyberTimon:
Thanks a lot! Looks good to me. Please remove the toasts and remove all translation files. For now, we keep error's english-only. On import success, the new presets appear in the presets panel so an additonal toast isn't needed.

jhorakcz:
Thanks - toasts are gone and the en.json keys with them, so there are no locale files left in the diff. 

Worth flagging since [#1461](https://github.com/jhorakcz/RapidRAW/issues/1461) was now also done in the meantime - this overlaps it. What's here on top is one read/write of presets.json instead of one per file, and failure isolation: the merged loop awaits inside a shared try, so the first unparseable file aborts the rest of the batch silently.

CyberTimon:
Thanks, looks good to me now :)

Issue #1398, opened 2026-07-19
FEATURE: More Export formats and color spaces


sebpaxx:
### Feature Description

I know it might be a niche feature, but it would be useful to have more export formats (like DNG or Tiff with different bitrates) and other color spaces too, like CMYK for printing photos

### Similar feature requests

- [x] I have searched existing issues and discussions and didn't find a duplicate

CyberTimon:
TIFF can now be exported in 16-bit since this PR: https://github.com/CyberTimon/RapidRAW/pull/1466

Issue #120, opened 2025-07-20
Export PNG / TIFF 16 bits


twenster:
Hello,

I can't see if PNG and TIF are saved with 8 or 16 bits depth. What is the current state ?

It would be good in the future to have an option in the export panel to save either 8 or 16 bits.

Side question : what colourspace and bit depth are you working internally ? (ROMM/ProPhoto RGB / 32 , rec2020 / 32 , something else)

Thank you !

BinAlexme:
Hi! 

Please add support for HEIC formats.

CyberTimon:
@BinAlexme 
https://github.com/CyberTimon/RapidRAW/issues/66

@twenster 
Will try to add this soon. The shader performs all its internal calculations in Linear RGB with sRGB / Rec. 709 primaries, using 32-bit floating-point precision.

CyberTimon:
TIFF can now be exported in 16-bit since this PR: https://github.com/CyberTimon/RapidRAW/pull/1466

PR #1466, opened 2026-08-03
feat: add true 8-bit and 16-bit TIFF export

dimafa:
## Description

Add a TIFF-only 8-bit/16-bit export control and preserve high-precision renderer data through GPU readback so 16-bit TIFF exports contain high-precision RGB16 samples instead of promoted RGB8 values.

The existing TIFF encoder always requested RGB16, but the production renderer returned `ImageRgba8`. Converting that result to RGB16 only multiplied 8-bit samples by 257, producing a nominally 16-bit TIFF without additional image precision.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Documentation update
- [x] UI/UX improvement

## Changes Made

- Added a TIFF-only bit-depth selector with localized 8-bit and 16-bit choices; 16-bit is the default.
- Persisted TIFF depth in export presets while defaulting legacy presets and missing Rust settings to 16-bit.
- Added `--tiff-bit-depth 8|16` to headless export with strict validation and a 16-bit default.
- Added an `rgba16float` tiled GPU output/readback path used only for 16-bit TIFF export.
- Kept the existing sRGB transfer behavior in the shared shader and disabled only the RGB8 quantization dither for high-precision output.
- Added genuine RGB8 and RGB16 TIFF encoding without an 8-bit intermediate in the 16-bit path.
- Audited resize and watermark post-processing so 16-bit images stay high precision.
- Added focused tests for CLI parsing, serde compatibility, TIFF encoding, resize precision, and shader behavior.

## Screenshots/Videos
16-bit-tiff

## Testing

**Test Configuration:**

- **OS:** macOS 26.4
- **Hardware:** Apple M5 Pro, 24 GB RAM

**Native and headless validation:**

- Verified the TIFF-only selector, 16-bit default, 8-bit selection, size estimate, and hidden state for JPEG in the native Tauri application.
- Exported the same 6000 × 4000 ARW as direct RGB8 and RGB16 TIFFs using the same adjustments.
- RGB8 output: unsigned 8/8/8-bit RGB, 72,000,786 bytes.
- RGB16 output: unsigned 16/16/16-bit RGB, 144,001,354 bytes.
- RGB16 unique values: R 5,914; G 5,821; B 5,716.
- RGB16 values not divisible by 257: R 98.68%; G 99.45%; B 99.13%.
- Down-quantized RGB16 matched direct RGB8 within one 8-bit step for 100% of samples.
- Repeated RGB8 and RGB16 renders were byte-identical; an omitted CLI depth matched explicit 16-bit output.
- Invalid `--tiff-bit-depth 12` was rejected with process exit code 2.

## Checklist

## Additional Notes

- Related to #120 and #1398. PNG remains outside this PR's scope, while #1398 also requests additional formats and color spaces; both issues therefore remain only partially addressed.
- PR #1395 adds overlapping internal high-precision GPU groundwork for future JPEG XL work, but intentionally does not wire it into export. This PR is independently based on current `main` and implements the TIFF product path end to end.
- Shader output at the readback boundary is already sRGB-encoded (sRGB/Rec. 709 primaries), so the high-precision path preserves those code values and does not apply a second transfer function.
- TIFF exports remain uncompressed and do not embed an ICC profile, matching current behavior.
- The lazy high-precision tile is about 40.5 MiB versus 20.25 MiB for the RGB8 tile. A 6000 × 4000 RGBA16 CPU readback is about 183 MiB versus 91.6 MiB for RGBA8.

## AI Disclaimer:


Redo11:
Real issue, please fix

akhater:
Good catch that the encoder was never the broken part — `DynamicImage::to_rgb16` already quantises f32 correctly, it was being handed 8-bit data. That narrows the fix a lot, and the `"tif" | "tiff"` match in here quietly closes a second bug: `.tif` exports were failing outright with "Unsupported file format".

Taken into [Argentum](https://github.com/akhater/Argentum) — our RapidRAW fork, and as sharp as a prime — with the encoder arm borrowed and marked so we drop our copy when this lands: [`30ec639`](https://github.com/akhater/Argentum/commit/30ec6394). Shipped in [26.38.1](https://github.com/akhater/Argentum/releases/tag/26.38.1).

One measurement that might be worth having here. We put a full-range ramp through the real pipeline with no adjustments, and it comes back sitting exactly on the half-float grid — **1024 distinct levels per stop of exposure, and no more**. Sweeping the ramp density confirms it: 2048 inputs give 2048 levels, 4096 give 3073, 8192 give 4097, each doubling adding one more stop's worth of 1024.

That's ~11 significant bits. The `rgba32float` render target isn't the ceiling — **the upload is**, since the image still reaches the GPU as `Rgba16Float` exactly as it does for previews. So this is a real and large improvement over shipping 8-bit data in a 16-bit container, but the last of the gap is on the input side, not the output. Happy to share the test if it's useful.

Nice work.

CyberTimon:
Thanks a lot for working on this PR! Looks good to me, will merge now 💯

PR #1478, opened 2026-08-06
Add side panel visibility toggles

dimafa:
## Description

Adds VS Code-style controls for completely hiding and restoring the left and right side panels. Panel visibility is now independent of panel width, so each panel returns to its exact previous size instead of reopening at a hard-coded width.

The default shortcuts are `Cmd+B` for the left panel and `Cmd+Shift+B` for the right panel on macOS (`Ctrl+B` and `Ctrl+Shift+B` elsewhere). Both remain configurable in Settings. These match VS Code shortcuts.

## Type of Change

- [x] New feature
- [x] UI/UX improvement

## Changes Made

- Add left/right panel toggle buttons to the custom title bar.
- Add configurable left/right panel keyboard actions and localized labels.
- Store visibility separately from width so buttons, shortcuts, and drag-to-collapse restore the exact previous panel size.
- Persist visibility in workspace settings and migrate legacy collapsed-width state.
- Reveal a panel automatically when an action selects content assigned to that side.
- Hide collapsed panel contents from pointer and accessibility interaction.

## Screenshots/Videos

LeftPanelHidden

BothPanelsHidden

## Testing

**Test Configuration:**

- **OS:** macOS 26.4
- **Hardware:** Apple M5 Pro

## Checklist

## Additional Notes

The previous implementation represented a collapsed panel by overwriting its width with 48 px. Because that discarded the prior width, it could not restore the user's chosen size. This change keeps width and visibility as separate state.

## AI Disclaimer:

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -28,7 +28,7 @@ import LibraryView from './components/views/LibraryView';
 
 import { ContextMenuProvider } from './context/ContextMenuContext';
 import { useSettingsStore } from './store/useSettingsStore';
-import { useUIStore } from './store/useUIStore';
+import { MIN_SIDE_PANEL_WIDTH, useUIStore } from './store/useUIStore';
 import { useLibraryStore } from './store/useLibraryStore';
 import { useEditorStore } from './store/useEditorStore';
 import { useProcessStore } from './store/useProcessStore';
@@ -116,6 +116,8 @@ function App() {
     isLibraryExportPanelVisible,
     leftPanelWidth,
     rightPanelWidth,
+    leftPanelVisible,
+    rightPanelVisible,
     compactEditorPanelHeightOverride,
     activePanel,
     activeLayoutDragItem,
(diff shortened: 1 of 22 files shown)


CyberTimon:
Hi @dimafa,

While reworking the implementation and storage for the panel width and visibility, I decided to directly implement the side panel visibility toggles (along with keybinds) as part of these changes: https://github.com/CyberTimon/RapidRAW/commit/491e4cf9f949f4db887b187ad4792f1e2b4ae432 & https://github.com/CyberTimon/RapidRAW/commit/8731af634762c39077e2ead470112f016a1e5382

Additionally, I added the toggles to the bottom bar to avoid cluttering the title bar, as title bar behavior isn't consistent across all operating systems.

Thanks!

dimafa:
@CyberTimon, the change looks great, thank you. 
Couple of suggestions:
1. If user uses key combos for Folder, Metadata or Export and left panel is Hidden, it should be displayed. Same for right panel. 
2. The default key combo is reversed compared to VS Code. (Ctrl-B is left and Ctrl-Shift-B is right) for coders, this behaviour is ingrained in muscle memory :-)

CyberTimon:
> If user uses key combos for Folder, Metadata or Export and left panel is Hidden, it should be displayed. Same for right panel.

Agreed. This can be tracked for the next release.

> The default key combo is reversed compared to VS Code. (Ctrl-B is left and Ctrl-Shift-B is right) for coders, this behaviour is ingrained in muscle memory :-)

This is intended. Since most photographers aren't using VS Code shortcuts on a daily basis, they aren't aware of this convention. The right panel is arguably the most used, which is why we use Ctrl+B for the right panel instead of the left.

Issue #1661, opened 2026-09-01
FEATURE: Display crop dimensions in pixels in the Crop & Transform panel


chuckhenrich:
### Feature Description

When the crop tool is active, the Crop & Transform panel should display the current crop region's dimensions in output image pixels, updated live as the user drags crop handles.

### Additional Context

**Proposed behavior**

- A pixel readout (e.g. 3,840 × 2,400 px) is shown in the Crop & Transform panel, between the Aspect Ratio section and the Rotation section.
- The readout is visible any time the crop tool is active — not only while dragging.
- When no crop has been applied, it shows the full oriented image dimensions.
- Dimensions are in output image pixels, not screen pixels. Screen pixel counts are window-size-dependent and not meaningful to the photographer.
- The readout updates in real time as crop handles are dragged, giving immediate feedback on the resulting image size.

**Why this is useful**

Photographers frequently need to know the pixel dimensions of a crop before exporting, for example, to confirm an image meets a minimum resolution for print, or to hit a specific pixel target for web delivery. Currently there is no way to see this without exporting first. Other RAW editors (e.g. Lightroom, Capture One, RawTherapee) surface this information directly in the crop tool.

**Prior art in this repo**

Discussion #501 included a related suggestion: "When a photo is cropped, display the new cropped size somewhere." This issue formalizes that request with a specific placement and behavior.

### Similar feature requests

- [x] I have searched existing issues and discussions and didn't find a duplicate

peterfrompoland:
I would love this feature of seeing what dimensions the photo is being cropped to. Otherwise the user has to export the image to see what the crop dimensions came out to, unless I'm missing some info panel. I just used RapidRAW for the 1st time and it's so user friendly and well designed! This was the only issue for me. I don't know if this is the place for this comment, but I just wanted to 'upvote' this feature.

PR #1662, opened 2026-09-01
feat(crop): display live crop dimensions in pixels in Crop & Transfor…

chuckhenrich:
## Description

Implements the feature requested in #1661.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made

- `src/store/useEditorStore.ts`: Added `liveCropPixels: { width: number; height: number } | null` to editor state, initialized to `null`.
- `src/components/panel/Editor.tsx`: In `handleCropChange`, added a `setLive()` helper that writes the current crop dimensions in image pixels to `liveCropPixels` on every drag update. Clears `liveCropPixels` back to `null` in `handleCropComplete` when the drag is released.
-` src/components/panel/right/CropPanel.tsx`: Added a pixel readout between the Aspect Ratio section and the Rotation section. Displays `liveCropPixels` during an active drag, falling back to `adjustments.crop` at rest, and the full oriented image dimensions when no crop has been applied. Handles both `px` and `%` crop units. Uses `toLocaleString()` for locale-appropriate thousands separators.

## Testing

**Test Configuration:**

- **OS:** Ubuntu 26.04
- **Hardware:** Intel 16-thread CPU, NVIDIA GeForce RTX 2060 Rev. A

## Checklist

## AI Disclaimer:

### DIFF
--- a/src/components/panel/Editor.tsx
+++ b/src/components/panel/Editor.tsx
@@ -1715,6 +1715,10 @@ export default function Editor({ onBackToLibrary, onContextMenu, onImageSelect, 
       const H = isSwapped ? selectedImage.width : selectedImage.height;
       const rotation = liveRotation !== null && liveRotation !== undefined ? liveRotation : adjustments.rotation || 0;
 
+      const setLive = (pc: PercentCrop) => {
+        setEditor({ liveCropPixels: { width: Math.round((pc.width / 100) * W), height: Math.round((pc.height / 100) * H) } });
+      };
+
       const MIN_CROP_PX = 64;
       const minPctW = (MIN_CROP_PX / W) * 100;
       const minPctH = (MIN_CROP_PX / H) * 100;
@@ -1734,12 +1738,15 @@ export default function Editor({ onBackToLibrary, onContextMenu, onImageSelect, 
       if (checkCropValid(toPixel(percentCrop), W, H, rotation)) {
         setCrop(percentCrop);
         lastValidCropRef.current = percentCrop;
+        setLive(percentCrop);
+
         return;
       }
 
       if (!lastValidCropRef.current) {
         setCrop(percentCrop);
         lastValidCropRef.current = percentCrop;
+        setLive(percentCrop);
         return;
       }
 
@@ -1826,6 +1833,7 @@ export default function Editor({ onBackToLibrary, onContextMenu, onImageSelect, 
 
         setCrop(finalCrop);
         lastValidCropRef.current = finalCrop;
+        setLive(finalCrop);
         return;
       }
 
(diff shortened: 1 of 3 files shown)


CyberTimon:
@chuckhenrich Thanks for the PR. I like the idea :)

Do you have other locations to display this information in mind? Don't just implement it somewhere else, let's discuss first. I think it looks out of place at that spot the crop panel.

chuckhenrich:
Sorry! I didn't mean to be presumptuous, I just put the dimensions there as a suggestion.

Another option is to display the dimensions live over the crop area. I use the Gnome extension Window Dimensions (https://github.com/brainicorn/windim). It displays the dimensions live as you drag window handles.

https://github.com/user-attachments/assets/8f97eebc-f424-448c-aa3a-9b86f1699335

What do you think about that kind of approach for the crop tool?

CyberTimon:
@chuckhenrich 

Sorry for the late reply!
And no worries at all :) the only reason I mentioned discussing it first is that a lot of AI agents tend to immediately rewrite the whole implementation as soon as I say I don't like something, so I just wanted to make sure we're on the same page before you spent any time on it.
Regarding the overlay, I'm still not 100% sure about displaying it in the middle of the crop area, as it might get in the way while fine-tuning the framing.
What about displaying it in the editor toolbar next to the actual image resolution, but only while the crop tool is active? What do you think?

chuckhenrich:
No worries, I know you're busy!

Absolutely yes, crop dimensions should be visible only when the crop panel is active. 

I think giving the user immediate, live feedback about the crop dimensions as they drag handles is essential, as well as the ability to enter specific dimensions "on the fly" instead of creating a custom aspect ratio. 

If that's the plan, then there's pros and cons to putting the crop dimensions in the editor toolbar next to the image resolution (I'm guessing you mean up top of the window where the filename and image dimensions are displayed?):

image

Pros: that serves as a heads up display where the user's eyes are naturally going to travel the least distance to get the numeric feedback. Showing those numbers (width/height of the crop) in the crop panel makes the user look side to side to get that feedback.

Con: the user couldn't specify specific dimension numbers in width and height fields if those fields are in the editor toolbar/titlebar unless those fields can be made editable in that toolbar/titlebar. 

Is it possible to have editable fields in the editor toolbar/titlebar alongside the file name and dimensions?

CyberTimon:
Okay no then I think it's best to keep it on the right in the crop panel (let's not make editor toolbar info editable). Maybe we can take some design inspiration from the ai panel's ai status pill, which could also fit into the crop panel at the top to display the dimensions?
image

What do you think?

Issue #990, opened 2026-04-08
BUG: Distortion correction is not linear, but warps at the ends


metboef:
### What happened?

when forcing distortion (over the lens value) the image horizon is not linear, but is like a wave: left up, then down, then up at the center, then down, then up againg at the right end.

### What did you expect to happen?

when correcting any distortion the horizon should be a straight line (when done with the right value), instead is not possible to get a straight horizon

### Operating System

Windows 11

### RapidRAW Version

1.5.1. & 1.5.3

### Acknowledgement

- [x] I have verified that my camera and lenses are supported by the external libraries listed above, or my issue is unrelated to camera/lens support.
- [x] I have searched all existing issues, both open and closed, and did not find a duplicate.

PR #1681, opened 2026-09-04
Fix lensfun-based distortion correction

duddel:
## Description

Applying lensfun data to my lens (Nikkor Z 24-70mm f/4 S) yields, what looks like, overcompensation, especially in image center.

For comparison, I checked darktable 5.6.0 against RapidRAW 1.6.2. Both use the exact same lensfun value for my lens, but the first tool generates matching results to the in-camera jpeg processing. I think darktable uses the lensfun reference implementation.

After digging into the code, I found 3 issues.

(1)
`lens_dist_amt` has an arbitrary scale of `2.5`. While this makes sense to increase sensitivity in certain cases, here it does not. The `lens_distortion_amount` from GUI is scaled to `[0, 2]`, default `1.0` (slider value `100`). To get the vanilla correction factor, the slider value would need to be `40`.

(2)
The `half_diagonal` is used for the normalized `radius` (`ptlens`, `poly3`, `poly5`), but it must be the "half short edge of the image". Vignetting still uses the half diagonal.

(3)
The `poly3` model defines only `k1`, which boils down to `ptlens` model with `a=0, b=k1, c=0`. (Not relevant for my observation, but still an issue)

I hope I found all occurrences where this is relevant.

This PR should fix most of the issue, but we might also need to consider image aspect ratio vs. calibration aspect ratio.

Also, not sure if TCA also needs a treatment.

References: lensfun code and docs.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made

(1)
Removed the `2.5` scaling factor

(2)
Using "half short edge of the image" now

(3)
added distinct `poly3` case in `extract_dist_params()`

## Screenshots/Videos

Can supply example image if required.

## Testing

- [ ] I haven't added any automated tests to the code because the codebase currently lacks a test suite.

**Test Configuration:**

- **OS:** Ubuntu 24.04
- **Hardware:** Ryzen 5000 series, Nvidia RTX 3060

## Checklist

- [ ] My changes generate no new warnings or errors

## AI Disclaimer:

- [ ] This PR is mostly AI-generated but edited/merged together by a human
- [X] This PR was handwritten with AI assistance (spell check, logic suggestions, error resolving)

### DIFF
--- a/src-tauri/src/image_processing.rs
+++ b/src-tauri/src/image_processing.rs
@@ -602,13 +602,13 @@ fn solve_generic_distortion_inv(r_target: f64, k_scaled: f64) -> f64 {
 fn compute_lens_auto_crop_scale(params: &GeometryParams, width: f32, height: f32) -> f64 {
     let cx = (width / 2.0) as f64;
     let cy = (height / 2.0) as f64;
-    let half_diagonal = (cx * cx + cy * cy).sqrt();
+    let dist_norm = cx.min(cy);
     let max_radius_sq_inv = 1.0 / (cx * cx + cy * cy);
 
     let lk1 = params.lens_dist_k1 as f64;
     let lk2 = params.lens_dist_k2 as f64;
     let lk3 = params.lens_dist_k3 as f64;
-    let lens_dist_amt = (params.lens_distortion_amount as f64) * 2.5;
+    let lens_dist_amt = params.lens_distortion_amount as f64;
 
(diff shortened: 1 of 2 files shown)


duddel:
Sounds similar to #990.

PR #1686, opened 2026-09-05
Detect fixed lenses through the Lensfun mount

beneedict:
## Problem

A camera with a fixed lens writes no `LensModel` EXIF tag. Panasonic compacts, many Canon compacts and most phone cameras leave it out.

Auto detection stops at that point and reports no result:

```rust
// file_management.rs
let exif_model = exif.get("LensModel").map(|s| s.as_str()).unwrap_or("");
```

```tsx
// CropPanel.tsx
if (!exifMaker || !exifModel) {
  setDetectionStatus('not_found');
  return;
}
```

Lensfun knows these lenses. It gives such a camera its own mount, and exactly one lens uses that mount. The information is there, RapidRAW just does not use it.

## Change

Add `find_lens_by_camera_mount`. When `LensModel` is absent, or when the fuzzy search finds nothing, look up the mount of the camera model and take the lens on that mount.

**The fallback applies only when the mount is unambiguous.** If a mount has more than one lens, the camera has interchangeable lenses, and the search stops. Otherwise a Canon EOS body would get an arbitrary EF lens.

The bundled database supports this rule:

| Lenses on the camera mount | Cameras | Meaning |
|---|---|---|
| exactly 1 | 366 | fixed lens, the match is certain |
| more than 1 | 649 | interchangeable lens, no match possible |
| none | 35 | no data |

## Example

A Panasonic DMC-LX15 now resolves to Leica `DMC-LX10 & compatibles (Standard)` through the mount `panasonicLX10`. Before the change auto detection reported nothing for this camera.

## Notes

- `find_best_lens_match` and the `autodetect_lens` command take one more argument, the EXIF tag `Model`. The frontend passes it.
- The fuzzy search over `LensModel` keeps priority. A file that worked before behaves exactly as before.
- The comparison of maker and model is exact and ignores upper and lower case. It checks every language variant of a Lensfun name.
- Batch processing uses the same path through `resolve_lens_params_in_adjustments`.

### DIFF
--- a/src-tauri/src/file_management.rs
+++ b/src-tauri/src/file_management.rs
@@ -2078,9 +2078,15 @@ pub fn resolve_lens_params_in_adjustments(
             if let Some(exif) = exif_data {
                 let exif_maker = exif.get("Make").map(|s| s.as_str()).unwrap_or("");
                 let exif_model = exif.get("LensModel").map(|s| s.as_str()).unwrap_or("");
+                let exif_camera_model = exif.get("Model").map(|s| s.as_str()).unwrap_or("");
                 if let Some(db) = lens_db {
                     if let Some((detected_maker, detected_model)) =
-                        crate::lens_correction::find_best_lens_match(db, exif_maker, exif_model)
+                        crate::lens_correction::find_best_lens_match(
+                            db,
+                            exif_maker,
+                            exif_model,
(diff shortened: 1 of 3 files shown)


CyberTimon:
Thanks, LGTM!

PR #1692, opened 2026-09-05
fix(macOS): register RAW file UTIs for file associations

seltom73:
## Summary

Fixes macOS RAW file associations by explicitly registering stable RAW UTIs in Tauri's `fileAssociations`.

Previously, RapidRAW listed many RAW extensions in a single association, but the generated macOS `Info.plist` only contained `LSItemContentTypes` for a few standard image formats. As a result, macOS did not consistently list RapidRAW as a recommended application for RAW files such as CR3 and ARW.

This change:
- adds explicit `contentTypes` for RAW formats with stable macOS UTIs,
- keeps extensions without stable UTIs associated by extension only,
- leaves standard image formats in the generic image association.

## Tested

Verified on macOS that the generated `Info.plist` now contains the expected `LSItemContentTypes`, including:

- `com.canon.cr3-raw-image`
- `com.sony.arw-raw-image`
- `com.adobe.raw-image`
- `com.nikon.raw-image`

The bundled app can then be registered with LaunchServices and appears correctly as a recommended application for supported RAW files.

### DIFF
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -39,39 +39,177 @@
     "category": "Photography",
     "fileAssociations": [
       {
+        "ext": ["dng"],
+        "contentTypes": ["com.adobe.raw-image"],
+        "name": "DNG RAW Image",
+        "description": "DNG RAW image",
+        "role": "Editor"
+      },
+      {
+        "ext": ["crw"],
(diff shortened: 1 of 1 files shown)


CyberTimon:
Great, thanks a lot!

PR #1715, opened 2026-09-08
Make sure cached EXIF ends up in sidecar file, so lens correction can work

### LINKED ISSUE #1706: BUG: Auto lens correction lost after switching images (sidecar without exif block is stripped on save)
### What happened?

After clicking **Auto** in Lens Correction, the lens is detected and the correction is applied. As soon as I switch to another image and come back, the correction is gone: the panel shows "Waiting for Auto-Detect", the three toggles are greyed out, and nothing is corrected.

The `.rrdata` sidecar written by RapidRAW contains `lensCorrectionMode: "auto"` and the three amount sliders, but **no** `lensMaker`, `lensModel` or `lensDistortionParams`. The detected lens is stripped on every save.

The batch action "Auto Lens Correction" in the library context menu (`apply_auto_lens_correction_to_paths`) is affected the same way and silently does nothing for these folders.

The detected lens and its coefficients should persist in the sidecar, so the correction is still applied when the image is reopened.

1. Open a folder of RAW files that has never been opened in RapidRAW before (no `.rrdata` sidecars exist yet). Let the thumbnails generate.
2. Open an image → Lens Correction → **Auto**. The lens is detected and applied.
3. Switch to another image, then back to the first one.
4. Panel shows "Waiting for Auto-Detect", no correction. Open the `.rrdata`: no `lensMaker` / `lensModel` / `lensDistortionParams`, and no `exif` block.

### Operating System

Windows 11

### RapidRAW Version

v1.6.3 (the relevant code is unchanged on `main` as of 2026-09-06)

### Additional information

**Root cause** (traced through the source):

1. **EXIF is cached app-side, not in the sidecar.** On first browse there is no `.rrdata`, so `read_exif_data` / `persist_exif_if_missing` in `exif_processing.rs` store the EXIF in `<app_cache_dir>/exif/<blake3(folder)>.json` via `save_exif_to_rrcache`.
2. **It never migrates into the sidecar.** Both functions return early when `read_rrexif_sidecar` finds the cache entry, so once a `.rrdata` is created later it never receives an `exif` block. In my test folder 0 of 47 sidecars have one.
3. **The save path only looks at the sidecar's EXIF.** `save_metadata_and_update_thumbnail` in `file_management.rs` does

   ```rust
   let mut metadata = crate::exif_processing::load_sidecar(&sidecar_path);
   ...
   resolve_lens_params_in_adjustments(&mut final_adjustments, &metadata.exif, lens_db_guard.as_deref());
   ```

   and `resolve_lens_params_in_adjustments`, in auto mode with `exif_data == None`, runs `map.remove("lensMaker"); map.remove("lensModel");` and then removes `lensDistortionParams` because no valid lens remains. The frontend detection succeeds (it uses the live EXIF of the loaded image), but the persisted result is stripped.

Side effects of the same root cause:

- `apply_auto_lens_correction_to_paths` and `apply_adjustments_to_paths` pass `existing_metadata.exif` to the same resolver, so batch auto lens correction is a no-op for these folders.
- In **manual** mode the lens survives, but without EXIF `focal_length` defaults to `50.0`, so a zoom lens gets the 50 mm coefficients persisted regardless of the real focal length.

**Suggested fix**

Fall back to the cached EXIF when the sidecar has none (in `save_metadata_and_update_thumbnail` and the two batch commands), and persist it so later loads are consistent:

```rust
let mut metadata = crate::exif_processing::load_sidecar(&sidecar_path);
if metadata.exif.is_none() {
    metadata.exif = crate::exif_processing::read_rrexif_sidecar(&source_path);
}
```

Alternatively, let `read_exif_data` / `persist_exif_if_missing` write the cached EXIF into the primary sidecar whenever the sidecar exists but lacks `exif`, instead of returning early on a cache hit.

**Workaround I used**

Copied the folder's cached EXIF entries from `<app_cache_dir>/exif/*.json` into each `.rrdata` as the `exif` field. After that, Auto persists correctly and the batch "Auto Lens Correction" action works.

### Acknowledgement

duddel:
## Description

Fixes #1706.

New helper `load_sidecar_with_exif()` falls back to the cached EXIF when the sidecar has none, so `resolve_lens_params_in_adjustments()` sees it and no longer strips the detected lens.

sidecar files, written after that, will contain the entire EXIF block from cache from then on.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Documentation update
- [ ] UI/UX improvement

## Changes Made

_see above_

## Testing

- [ ] I haven't added any automated tests to the code because the codebase currently lacks a test suite.

**Test Configuration:**

- **OS:** Ubuntu 24.04
- **Hardware:** Ryzen 5000 series, Nvidia RTX 3060

## Checklist

- [ ] My changes generate no new warnings or errors

## AI Disclaimer:

- [X] This PR is mostly AI-generated but edited/merged together by a human

### DIFF
--- a/src-tauri/src/exif_processing.rs
+++ b/src-tauri/src/exif_processing.rs
@@ -244,6 +244,14 @@ pub fn load_sidecar(sidecar_path: &Path) -> ImageMetadata {
     meta
 }
 
+pub fn load_sidecar_with_exif(sidecar_path: &Path, source_path: &Path) -> ImageMetadata {
+    let mut meta = load_sidecar(sidecar_path);
+    if meta.exif.is_none() {
+        meta.exif = read_rrexif_sidecar(source_path);
+    }
+    meta
+}
+
 fn to_ur64(val: &exif::Rational) -> uR64 {
     uR64 {
         nominator: val.num,
--- a/src-tauri/src/file_management.rs
+++ b/src-tauri/src/file_management.rs
@@ -2527,7 +2527,7 @@ pub fn save_metadata_and_update_thumbnail(
 ) -> Result<(), String> {
     let (source_path, sidecar_path) = parse_virtual_path(&path);
 
-    let mut metadata = crate::exif_processing::load_sidecar(&sidecar_path);
+    let mut metadata = crate::exif_processing::load_sidecar_with_exif(&sidecar_path, &source_path);
 
     let mut final_adjustments = adjustments;
     {
@@ -2636,9 +2636,10 @@ pub async fn apply_adjustments_to_paths(
             .clone();
 
         paths.par_iter().for_each(|path| {
-            let (_, sidecar_path) = parse_virtual_path(path);
+            let (source_path, sidecar_path) = parse_virtual_path(path);
 
-            let mut existing_metadata = crate::exif_processing::load_sidecar(&sidecar_path);
+            let mut existing_metadata =
+                crate::exif_processing::load_sidecar_with_exif(&sidecar_path, &source_path);
 
             let mut new_adjustments = existing_metadata.adjustments;
             if new_adjustments.is_null() {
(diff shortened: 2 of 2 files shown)


CyberTimon:
Thanks a lot for the fix. LGTM!

PR #1758, opened 2026-09-21
feat: hold Ctrl to resize the crop from its center

lalibertemarc:
## Description

Holding Ctrl (Cmd on macOS) while dragging a crop handle now resizes the crop around its center: the side opposite the handle you drag moves by the same amount, so the center of the frame stays put.

I know about the existing Ctrl + scroll crop zoom, and this does not replace or change it. I added this because holding a modifier while dragging a handle is the gesture most people already have in their fingers from the dominant commercial photo editor, and because it is more precise in practice: a drag is continuous and lets you land on an exact framing, while the wheel moves in fixed 5% steps. Ctrl + scroll is also anchored to the cursor and slides the crop back inside the image when it reaches an edge, which shifts the framing you were aiming for. This gesture keeps the center fixed and simply stops growing at the edge.

Both gestures are useful for different things, so they now sit side by side under the same modifier: scroll for a quick rough resize, drag for the exact one.

## Type of Change

- [x] New feature

## Changes Made

`src/components/panel/Editor.tsx`

- On a pointer press on a crop handle, remember the crop as it was at that moment.
- In `handleCropChange`, when Ctrl is held during a handle drag, mirror react-image-crop's anchored resize around that remembered center. Pressing or releasing Ctrl mid-drag switches between centered and normal resizing.
- If the centered crop would leave the image (including when the image is rotated), it grows only until the first side reaches the edge, so the center never drifts.
- The remembered crop is cleared on pointer release and on any press that is not on a handle.

`src/utils/cropUtils.ts`

- Added `fitCropTowards`, a small helper that walks from one crop toward another and returns the largest version that still passes a validity check.
- This is the reason for the second file in the diff. That binary search already exists in several places (`zoomCrop`, `forceCropInBounds`, and three times in `handleCropChange`), so rather than adding another copy inside the new code path, I pulled the shape out into one generic helper. It is typed generically so it works with both the pixel crops in `cropUtils` and the percentage crops that react-image-crop hands back.
- I deliberately did not rewire the existing call sites. Doing so would change code paths that already work, including the merged Ctrl + scroll zoom, and that does not belong in this PR. If you would like them to use the helper as well, I am happy to do it here or in a follow-up.

Nothing else changed. No existing keybind or gesture was modified: Ctrl + scroll crop zoom, Ctrl + drag crop panning outside crop mode, and Ctrl + double-click reset all behave exactly as before.

## Screenshots/Videos

I can provide a short clip if needed

## Testing

Tested by hand: corner and edge handles, free and locked aspect ratios, rotated images, pressing and releasing Ctrl in the middle of a drag, and dragging outward until the crop hits the image bounds. Also re-checked the existing gestures listed above for regressions.

**Test Configuration:**

- **OS:** Windows 10 Home (19045)
- **Hardware:** AMD Ryzen 5 3600X, NVIDIA RTX 2070, 16 GB RAM

## Checklist

`npm run lint`, `npm run format:check` and `tsc --noEmit` report nothing new for the two changed files.

## Additional Notes

Happy to adjust the modifier if you would rather keep Ctrl purely for the existing crop pan and zoom. I avoided Alt because several Linux window managers swallow Alt + drag to move windows, and Shift because it usually means "keep the aspect ratio" in editors of this kind.

## AI Disclaimer:


CyberTimon:
Great PR! LGTM :)

